PR #1, opened 2021-08-11
Upload PLDM FileIO Spec v2

manojkiraneda:
Uploading the PLDM fileIO spec to that Inspur systems can be able to access it.


geissonator:
add to approvelist

PR #2, opened 2021-08-12
Add host_container_id sent from Host

lxwinspur:
Currently bmc normalizes the entity association PDRs that are sent
by Host, and bmc will generate a new container ID.

Since the new container ID is inconsistent with the container ID sent
by Host, we also need to temporarily save the host_container_id sent
by Host.

Tested: built pldm successfully and Unit Test passes.

Signed-off-by: George Liu <liuxiwei@inspur.com>
Change-Id: Ib564554b191d1ffe2007b831cb22b5cf9b67d57f


lxwinspur:
This is a test case

PR #4, opened 2021-08-13
pldmd: use async instead of event loop at start

manojkiraneda:
This commit changes the first command to async that pldmd sends
after starting to discover Host state.

Also uses GetPLDMVersion as the first command so that Host understands
bmc pldmd has restarted.

Change-Id: Icb556a89f3cc9564450cc344ba9e0e8b56421fd9
Signed-off-by: Sampa Misra <sampmisr@in.ibm.com>


manojkiraneda:
This PR is up streamed in the following commit:
https://gerrit.openbmc-project.xyz/c/openbmc/pldm/+/45671

sampmisr (approved):

PR #10, opened 2021-09-01
Fix the looping behaviour in the getstatesensorreading command

manojkiraneda:
This bug was actually caused by two host sensors that has the
exact same entity id, instance id, container id but being used
to implement different state sets.

The old host code had two sensors on processor module that implements
both stateset 1 & 11 that caused this regression and using this PR we
have improved the sensor parsing abilities so that we can compare the
state setId as well before deciding that a sensor is unique.


manojkiraneda:
upstream commit : https://gerrit.openbmc-project.xyz/c/openbmc/pldm/+/39796

sampmisr (approved):

PR #19, opened 2021-09-21
oem-ibm: Resource dump related fixes

jaypadath:
This commit include following changes:

1. Allow the empty string as resource dump selector
2. Avoid setting the resource dump dump explicitly as notify method
   takes care of the same. This code was causing error when the
   resource dump is getting initiated from host

Tested and verified that resource dump initiation from host is
getting completed without any error.

Signed-off-by: Jayashankar Padath <jayashankar.padath@in.ibm.com>
Change-Id: I58f51cb115d86ae22099d57474a1159d8952cefa


geissonator:
add to approvelist

sampmisr (approved):

dhruvibm (approved):

PR #24, opened 2021-09-30
oem: ibm: Assemble the image in a forked process

anoo1:
The assembling of the tarball image file is a long running process,
causing pldm to not reply to any other requests until it's done, causing
errors such as when the hypervisor tries to ping the bmc to determine if
it's running.

Move the assembling of the image to a forked process so that it runs in
the background. Move the assembling function to the code update class so
that it has access to the code update sensor to set it to error if an
failure occurs, since no parent process will wait for it.

Use double fork to avoid zombie tasks, this allows the grandchild to be
reparented to init.

Tested: Verified the hypervisor did not reset the BMC during an inband
update because it didn't receive a reply to its ping request.

Change-Id: I47d5fde1a1311e94d5b03f7fcf4c0ab15ba054e3

### DIFF
--- a/oem/ibm/libpldmresponder/inband_code_update.cpp
+++ b/oem/ibm/libpldmresponder/inband_code_update.cpp
@@ -523,48 +523,116 @@ int processCodeUpdateLid(const std::string& filePath)
     return PLDM_SUCCESS;
 }
 
-int assembleCodeUpdateImage()
+int CodeUpdate::assembleCodeUpdateImage()
 {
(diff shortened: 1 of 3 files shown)


anoo1:
Gerrit: https://gerrit.openbmc-project.xyz/c/openbmc/pldm/+/47248

rfrandse:
@sampmisr 
Per @anoo1 comment:
merging without review, @mzipse  was interested in a driver so we can test the functionality while the upstream reviews are in progress

PR #49, opened 2021-11-11
PLDM: Setting the slot power state to Off

Pavithrab7:
This commit sets the slot power state to off when the
host powers off. The host only powers on the slots and
the BMC needs to turn all the slots off when the host powers off.

Tested using the busctl commands.

Change-Id: Ie1fe384638a756f2657ea113e8f8aa5e9caace2b

### DIFF
--- a/host-bmc/host_pdr_handler.cpp
+++ b/host-bmc/host_pdr_handler.cpp
@@ -173,6 +173,54 @@ HostPDRHandler::HostPDRHandler(
                 }
             }
         });
+
+    chassisOffMatch = std::make_unique<sdbusplus::bus::match::match>(
+        pldm::utils::DBusHandler::getBus(),
+        propertiesChanged("/xyz/openbmc_project/state/chassis0",
+                          "xyz.openbmc_project.State.Chassis"),
+        [this](sdbusplus::message::message& msg) {
+            DbusChangedProps props{};
+            std::string intf;
+            msg.read(intf, props);
+            const auto itr = props.find("CurrentPowerState");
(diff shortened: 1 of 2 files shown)


Pavithrab7:
upstream Gerrit commit - https://gerrit.openbmc-project.xyz/c/openbmc/pldm/+/48678

sampmisr (approved):

host-bmc/host_pdr_handler.cpp
```diff
@@ -173,6 +173,54 @@ HostPDRHandler::HostPDRHandler(
                 }
             }
         });
+
+    chassisOffMatch = std::make_unique<sdbusplus::bus::match::match>(
+        pldm::utils::DBusHandler::getBus(),
+        propertiesChanged("/xyz/openbmc_project/state/chassis0",
+                          "xyz.openbmc_project.State.Chassis"),
```
manojkiraneda, line 180: does the chassis power solve all the cases...what about the host state ?

spinler: PHYP leaves the slots on on a power off, so they will stay on until the hardware is turned off, which is the chassis.  We could maybe convince ourselves that host state would also work, but leaving it at chassis should be fine too.

manojkiraneda: Ack

manojkiraneda (approved):

Pavithrab7:
Some other changes have gone in and this PR when merged can cause merge conflicts. So closing it and opening a new one.

PR #51, opened 2021-11-11
oem: Fix to remove extra system dump

jaypadath:
An unwanted system dump is being generated if the resource dump is
initiated from hypervisor with an empty vsp string. So the fix here
is to additionally check the resource dump progress status and block
sending the command to hyperviosr if the status is not in progress.

Signed-off-by: Jayashankar Padath <jayashankar.padath@in.ibm.com>
Change-Id: I25c1db3510d6f44b354881e4065aa148fe618464


dhruvibm (approved):

jaypadath:
Creating a new PR as facing some issue with this change

PR #53, opened 2021-11-12
Fix crashes and potential crashes

manojkiraneda:
- This PR would fix crashes in various paths like dump,
  COD license paths including the VMI certificate exchange.

- This PR would also fix the crash that happens when the host
  off signal is received.


manojkiraneda:
all these crashes are coming from the merged code in master.I would also be putting commits upstream fixing these.

sampmisr (approved):

PR #62, opened 2021-11-18
PLDM: Setting the slot power state to Off

Pavithrab7:
This commit sets the slot power state to off when the
host powers off. The host only powers on the slots and
the BMC needs to turn all the slots off when the host powers off.

Tested using the busctl commands.

Change-Id: Ie1fe384638a756f2657ea113e8f8aa5e9caace2b


Pavithrab7:
Upstream Gerrit commit - https://gerrit.openbmc-project.xyz/c/openbmc/pldm/+/48678

sampmisr (approved):

PR #72, opened 2021-12-06
Add orphan entities to the Entity Association PDR's

manojkiraneda:
It is not always a standard condition that we should have
a fru record set PDR, we can also have sensors/effecters
without having an associated FRU record set PDR.

This commit would help add these orphan sensors/effecters
entities into the entity association PDR's.

Fixes : SW536590 & SW536589

### DIFF
--- a/libpldmresponder/pdr_state_sensor.hpp
+++ b/libpldmresponder/pdr_state_sensor.hpp
@@ -26,7 +26,8 @@ static const Json empty{};
  */
 template <class DBusInterface, class Handler>
 void generateStateSensorPDR(const DBusInterface& dBusIntf, const Json& json,
-                            Handler& handler, pdr_utils::RepoInterface& repo)
+                            Handler& handler, pdr_utils::RepoInterface& repo,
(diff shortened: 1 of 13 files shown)


libpldmresponder/pdr_state_sensor.hpp
```diff
@@ -73,11 +74,11 @@ void generateStateSensorPDR(const DBusInterface& dBusIntf, const Json& json,
 
         pdr->terminus_handle = TERMINUS_HANDLE;
         pdr->sensor_id = handler.getNextSensorId();
-
+        auto& associatedEntityMap = handler.getAssociateEntityMap();
```
lxwinspur, line 77: move to L: 79

manojkiraneda: done.

lxwinspur (approved):

manojkiraneda:
upstream commit :  https://gerrit.openbmc-project.xyz/c/openbmc/pldm/+/49420

sampmisr (approved):

PR #79, opened 2021-12-13
Spawn a new thread for the PDR Generation

manojkiraneda:
PDR generation process is quite dbus intensive & in the systems
with more dbus latency , it is observed that PLDM is blocked
while doing these dbus calls & as these are blocking calls pldm
is blocked & was not responding to host which untimately led to
the PHYP initiating the BMC reboot.

Although this does not completely fix the issue, as this only
unblocks the pldm daemon just for the PDR generation process,
This is just a temperory fix, the more concrete and future proof
solution is to switch to sdbusplus::asio infrastructure.

Tested By:

1. Power off & power on in loop - PASS
https://sys-openbmc-dev-jenkins.swg-devops.com/job/OpenBMC-Dev/job/test/job/continuous-integration/job/ebmc/job/power_on_off_loop/5/console

Suggested-by: Andrew Jeffery <andrew@aj.id.au>

### DIFF
--- a/libpldmresponder/platform.cpp
+++ b/libpldmresponder/platform.cpp
@@ -156,47 +156,8 @@ void Handler::generate(const pldm::utils::DBusHandler& dBusIntf,
     }
 }
 
-Response Handler::getPDR(const pldm_msg* request, size_t payloadLength)
+Response Handler::sendPDRs(const pldm_msg* request, size_t payloadLength)
 {
-    if (hostPDRHandler)
-    {
-        if (hostPDRHandler->isHostUp() && oemPlatformHandler != nullptr)
-        {
-            auto rc = oemPlatformHandler->checkBMCState();
-            if (rc != PLDM_SUCCESS)
-            {
-                return ccOnlyResponse(request, PLDM_ERROR_NOT_READY);
-            }
-        }
-    }
-
-    // Build FRU table if not built, since entity association PDR's
-    // are built when the FRU table is constructed.
-    if (fruHandler)
-    {
-        fruHandler->buildFRUTable();
-    }
-
-    if (!pdrCreated)
-    {
-        generateTerminusLocatorPDR(pdrRepo);
-        generate(*dBusIntf, pdrJsonsDir, pdrRepo, bmcEntityTree);
-        if (oemPlatformHandler != nullptr)
-        {
-            oemPlatformHandler->buildOEMPDR(pdrRepo);
-        }
-
-        pdrCreated = true;
-
-        if (dbusToPLDMEventHandler)
-        {
-            deferredGetPDREvent = std::make_unique<sdeventplus::source::Defer>(
-                event,
-                std::bind(std::mem_fn(&pldm::responder::platform::Handler::
-                                          _processPostGetPDRActions),
-                          this, std::placeholders::_1));
-        }
-    }
 
     Response response(sizeof(pldm_msg_hdr) + PLDM_GET_PDR_MIN_RESP_BYTES, 0);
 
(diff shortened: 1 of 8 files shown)


libpldmresponder/platform.cpp
```diff
@@ -261,6 +222,89 @@ Response Handler::getPDR(const pldm_msg* request, size_t payloadLength)
...
+                std::thread pdrGenThread(&Handler::_generatePDRs, this);
+
+                // detach the thread, so that the main thread wont be
+                // waiting for the response of the secondary thread
+                pdrGenThread.detach();
+
+                return ccOnlyResponse(request, PLDM_ERROR_NOT_READY);
+            }
+            else if (repoCreated.load() ==
+                     pldmRepoStatus::PLDM_REPO_BUILDING_IN_PROGRESS)
```
amboar, line 255: No sorry, this second load invokes a data race. This needs to be

```
else if (expected == pldmRepoStatus::PLDM_REPO_BUILDING_IN_PROGRESS)
```

libpldmresponder/platform.cpp
```diff
@@ -261,6 +222,89 @@ Response Handler::getPDR(const pldm_msg* request, size_t payloadLength)
...
+            {
+                return ccOnlyResponse(request, PLDM_ERROR_NOT_READY);
+            }
+            else if (repoCreated.load() ==
+                     pldmRepoStatus::PLDM_REPO_ALREADY_BUILT)
```
amboar, line 260: As above this load invokes a data race. It needs to be

```
else if (expected == pldmRepoStatus::PLDM_REPO_ALREADY_BUILT)
```

libpldmresponder/platform.cpp
```diff
@@ -261,6 +222,89 @@ Response Handler::getPDR(const pldm_msg* request, size_t payloadLength)
...
+                    deferredGetPDREvent =
+                        std::make_unique<sdeventplus::source::Defer>(
+                            event,
+                            std::bind(std::mem_fn(
+                                          &pldm::responder::platform::Handler::
+                                              _processPostGetPDRActions),
+                                      this, std::placeholders::_1));
+                }
+                return sendPDRs(request, payloadLength);
+            }
```
amboar, line 273: I think you need

```
} else {
    assert("Unreachable");
}
```
or something equivalent to make sure we have complete coverage of the enum.

libpldmresponder/platform.cpp
```diff
@@ -261,6 +222,89 @@ Response Handler::getPDR(const pldm_msg* request, size_t payloadLength)
...
+    }
+
+    // Host is off, this is a normal power on operation
+    // This is needed as hostboot does not handle the
+    // repository_in_progress status code and re-try
+    if (fruHandler)
+    {
+        fruHandler->buildFRUTable();
+    }
+    if (repoCreated == pldmRepoStatus::PLDM_REPO_NOT_BUILT)
```
amboar, line 284: You should do

```
if (repoCreated.compare_exchange_strong(expected, pldmRepoStatus::PLDM_REPO_BUILDING_IN_PROGRESS))
{
...
}
```
again here

amboar (changes requested):

libpldmresponder/platform.hpp
```diff
@@ -465,6 +493,9 @@ class Handler : public CmdHandler
     pldm::responder::oem_platform::Handler* oemPlatformHandler;
     sdeventplus::Event& event;
     std::string pdrJsonsDir;
+    std::atomic<pldmRepoStatus> repoCreated;
+
+    std::unique_ptr<sdeventplus::source::Defer> deferredPDRGeneration;
     bool pdrCreated;
```
amboar, line 499: This must be std::atomic<bool> pdrCreated as it's used concurrently across threads.

manojkiraneda:
we dont need this PR anymore (and maintainers are not ready to pull this in ) - as the host bumped up the timeouts.

PR #88, opened 2022-01-05
Fix: Set PCIeConfig Information for IBM Cards

manojkiraneda:
There is a bug in a code, where we ignore the PCIe Config
information that is sent by the PHYP for IBM Cable cards.

This PR would fix the issue.


geissonator (approved):
Manoj will update upstream review indicating this change is needed.

spinler (approved):

PR #133, opened 2022-02-16
Modify Dump offload to use Thread

raviteja-b:
Currently while offloading a Dump pldmd writes on unix sockets 1M chunks
When network is slow write to unix socket takes few mins which keeps pldmd
busy with write() calls in a loop

This is implemented as discussed 
1. PHYP sends the first 1MB chunk - BMC does the DMA operation, spin off's the thread 
  and then return a   PLDM_SUCCESS (after phyp gets the PLDM_SUCCESS they will unmap the TCE )
2. PHYP sends the second 1MB chuck - BMC would check the offload status and return PLDM_NOT_READY until the offload status is free

Tested By:
1. Offload system dump 
2.Offload system dump in slow networks speed

### DIFF
--- a/oem/ibm/libpldmresponder/file_io.cpp
+++ b/oem/ibm/libpldmresponder/file_io.cpp
@@ -19,6 +19,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <thread>
 
 namespace pldm
 {
@@ -58,11 +59,6 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
     {
         pageAlignedLength += pageSize;
     }
-
-    auto mmapCleanup = [pageAlignedLength](void* vgaMem) {
-        munmap(vgaMem, pageAlignedLength);
-    };
-
     int dmaFd = -1;
     int rc = 0;
     dmaFd = open(xdmaDev, O_RDWR);
@@ -77,20 +73,20 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
 
     pldm::utils::CustomFD xdmaFd(dmaFd);
 
-    void* vgaMem;
-    vgaMem =
+    vgaMemDump =
         mmap(nullptr, pageAlignedLength, PROT_READ, MAP_SHARED, xdmaFd(), 0);
-    if (MAP_FAILED == vgaMem)
+    if (MAP_FAILED == vgaMemDump)
     {
         rc = -errno;
         std::cerr
             << "transferHostDataToSocket : Failed to mmap the XDMA device, RC="
             << rc << "\n";
+        close(fd);
         return rc;
     }
 
-    std::unique_ptr<void, decltype(mmapCleanup)> vgaMemPtr(vgaMem, mmapCleanup);
-
+ std::cerr
+            << "transferHostDataToSocket :" << pageAlignedLength <<std::endl;
     AspeedXdmaOp xdmaOp;
     xdmaOp.upstream = 0;
     xdmaOp.hostAddr = address;
@@ -103,11 +99,13 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
         std::cerr
             << "transferHostDataToSocket : Failed to execute the DMA operation, RC="
             << rc << " ADDRESS=" << address << " LENGTH=" << length << "\n";
+        close(fd);
         return rc;
     }
-   std::cerr << "calling asyncThread \n"; 
-   std::future<void> asyncThread = std::async(std::launch::async, &writeToUnixSocket, fd, static_cast<const char*>(vgaMemPtr.get()), length);
-   std::cerr << "return after asyncThread \n"; 
+    std::cerr << "calling new thread \n"; 
+    std::thread dumpOffloadThread(writeToUnixSocket,fd, static_cast<const char*>(vgaMemDump), length);
+    dumpOffloadThread.detach();
+    std::cerr << "return after new Thread \n"; 
   return 0;
 }
 
--- a/oem/ibm/libpldmresponder/file_io.cpp
+++ b/oem/ibm/libpldmresponder/file_io.cpp
@@ -14,7 +14,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
-#include <future>
+
 #include <cstring>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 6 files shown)


oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -199,27 +199,79 @@ int DumpHandler::writeFromMemory(uint32_t, uint32_t length, uint64_t address,
     std::cout
         << "Start of DumpHandler::writeFromMemory transferFileDataToSocket"
         << std::endl;
-    return transferFileDataToSocket(DumpHandler::fd, length, address);
+
+    if (socketWriteStatus==Error)
```
sunharis, line 203: Suppose there is a case where 2 threads are running/waiting at a specific timeframe , when one thread hits error and we stop offloading the dump, how are we cleaning up the other thread?

oem/ibm/libpldmresponder/utils.hpp
```diff
@@ -113,6 +115,8 @@ bool checkFruPresence(const char* objPath);
  */
 void findPortObjects(const std::string& cardObjPath,
                      std::vector<std::string>& portObjects);
+
```
sunharis, line 118: extra lines?

oem/ibm/libpldmresponder/file_io_type_dump.cpp
raviteja-b: at a time only one thread allowed to run.
by any chance we hit case where two threads are there, as soon as first thread hit error, socket clean up and 2nd thread also hit error it will be cleaned up like first thread.
having critical section with mutex I don't think we hit this case

oem/ibm/libpldmresponder/file_io.cpp
```diff
@@ -106,18 +98,11 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
...
-    {
-        rc = -errno;
-        close(fd);
-        std::cerr << "transferHostDataToSocket : failure in closing socket"
-                  << std::endl;
-        return rc;
-    }
-    return 0;
+    std::cout << "calling new thread \n"; 
+    std::thread dumpOffloadThread(writeToUnixSocket,fd, static_cast<const char*>(vgaMemDump), length);
```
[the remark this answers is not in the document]
raviteja-b: no, we have already have checks before entering in this code

oem/ibm/libpldmresponder/file_io.cpp
```diff
@@ -77,21 +70,20 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
...
+        std::cout
             << "transferHostDataToSocket : Failed to mmap the XDMA device, RC="
             << rc << "\n";
         return rc;
     }
 
-    std::unique_ptr<void, decltype(mmapCleanup)> vgaMemPtr(vgaMem, mmapCleanup);
-
+    std::cout
+            << "transferHostDataToSocket :" << pageAlignedLength <<std::endl;
```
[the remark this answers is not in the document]
raviteja-b: sure. just wanted to have more traces for debugging any corner case.
we can remove later

oem/ibm/libpldmresponder/file_io.cpp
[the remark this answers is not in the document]
sunharis: but how do you ensure that the thread is spawned as expected ?

oem/ibm/libpldmresponder/file_io_type_dump.cpp
sunharis: There can be more than one thread at a time. But they will be waiting on the critical section for read/write operation. Yes, seems like they all can get cleaned up when the socketWriteStatus==Error happens. 

1. Who resets this socketWriteStatus to Free for the next dump offload?
2. At line 212 you will not get the socketInterface at the 2nd thread if its already cleaned up by the 1st thread. You will need to add a validity check before std::remove.
3. Does this PLDM_ERROR is sent to the phyp ? Will that not be a redundant message to phyp if its sent from all threads ?

raviteja-b: yeah we should avoid creating one more while one thread is running, otherwise we may end up issues you mentioned 
I will make have one thread at a time.

oem/ibm/libpldmresponder/file_io.cpp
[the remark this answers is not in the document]
raviteja-b: std:thread does not have any return type, its just create a object and invokes thread

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -200,23 +200,45 @@ int DumpHandler::writeFromMemory(uint32_t, uint32_t length, uint64_t address,
...
-          std::cout<< "Dump offload InProgress, sending Busy message" << std::endl;
-          return PLDM_ERROR_NOT_READY;
+        std::cout << "Dump offload InProgress, sending Busy message"
+                  << std::endl;
+        return PLDM_ERROR_NOT_READY;
     }
-    else if (socketWriteStatus==Completed )
+
+    auto rc = transferFileDataToSocket(DumpHandler::fd, length, address);
+    std::cerr << "wating for rc from transferFileDataToSocket " << rc
```
sunharis, line 225: why this is std::cerr

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -200,23 +200,45 @@ int DumpHandler::writeFromMemory(uint32_t, uint32_t length, uint64_t address,
         << "Start of DumpHandler::writeFromMemory transferFileDataToSocket"
         << std::endl;
 
-    if (socketWriteStatus==Error)
+    if (socketWriteStatus == Error || DumpHandler::fd == -1)
```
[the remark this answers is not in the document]
raviteja-b: its not redundant, there can be case where fd gets cleaned up on error, we should not keep returning NOT_READY error

sunharis (commented):
I am OK with the changes done. Need pldm team to approve.
@raviteja-b I hope you will be removing the traces before merging the code

oem/ibm/libpldmresponder/file_io_type_dump.cpp
raviteja-b: removed

oem/ibm/libpldmresponder/file_io_type_dump.cpp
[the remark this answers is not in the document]
raviteja-b: yeah this check is not  necessary, will remove

sunharis (approved):

manojkiraneda (approved):

PR #147, opened 2022-02-23
oem: ibm: Set next boot side attribute

anoo1:
During a rename, set the next boot side to Perm (same as current) so
that in case of a failed update, PHYP knows to boot from the Perm side.
The next boot side would be set to Temp on a successful update (via the
Priority callback function).

Update the next boot side attribute (fw_boot_side) whenever the next
boot side value is set or changed so that in error scenarios, PHYP knows
which side to boot from.

Setting fw_boot_side attribute automatically calls the function to set
the next boot side because the attribute can be also be set externally
by PHYP, therefore check the requested next boot side already has the
highest priority, in which case there's no need to set the Priority.
Also checking the current priority allows to catch if the alternate
version may not exist because of a code update that failed or is in
progress. Once an image is uploaded, set the non running version to the
uploaded image since the current non running image will be deleted
during the update.

Tested: Verified the fw_boot_side attribute matched the information in
the persisted bootSide file. The Priority value was only set when needed
(the requested next boot side was different than the current next boot
side). During a rename, both current and next are set to Perm.

Change-Id: Ibf44f53769a3394826796b1d101fcc39bdd2a4ef

### DIFF
--- a/oem/ibm/libpldmresponder/inband_code_update.cpp
+++ b/oem/ibm/libpldmresponder/inband_code_update.cpp
@@ -329,96 +329,106 @@ void CodeUpdate::setVersions()
                             imageObjPath, "Activation", imageInterface);
                         const auto& imageProp = std::get<std::string>(propVal);
                         if (imageProp == "xyz.openbmc_project.Software."
-                                         "Activation.Activations.Ready" &&
-                            isCodeUpdateInProgress())
+                                         "Activation.Activations.Ready")
                         {
                             nonRunningVersion = path.str;
-                            newImageId = path.str;
-                            if (!imageActivationMatch)
+
+                            if (isCodeUpdateInProgress())
                             {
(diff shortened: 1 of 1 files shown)


anoo1:
Added a second commit to this PR to avoid merge conflicts. The 2nd commit removes the flag check for a Rename event. The change is mostly space formatting bu github doesn't show the diff very well, here is the high level overview:

Before:
```
if Activation value is Ready && update in progress:
    // inband update
    nonrunning = path;
else if Activation value is Ready && oob not in progress:
    nonrunning = path;
    oob = true;
    do rename;
```

With change:
```
if Activation value is Ready:
    nonrunning = path;
    if (update in progress)
        // inband update
    else
       // out of band, do Rename
```

sagisin (approved):

manojkiraneda (approved):

PR #164, opened 2022-03-07
oem: ibm: Fix BootSide string values

anoo1:
The nextBootSide and currBootSide should have a value of "T" or "P", but
the side bios attributes have values of "Temp" or "Perm". There were a
couple places where the value of the attributes were being assigned to
the boot side variables without making the translation. Fixed by
converting "Temp/Perm" into "T/P".

Tested: Verified pldm returned the correct lid data from the requested
side after a code update.

Change-Id: I95f4087bfe382cf468a294fb18fdc8309b5b0bbc


geissonator (approved):
In the essence of time (CI) and the criticality of this defect, I'm going to merge so we can get downstream CI going. We will not merge that downstream change until the maintainers get a chance to look this over.

PR #178, opened 2022-03-16
PLDM: APR changes in inband CU path

sagisin:
This commit adds code to let the host know
that the boot initiator is host in the inband
codeupdate reboot path.

Defect:SW541646

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
@@ -1188,6 +1188,9 @@ void pldm::responder::oem_ibm_platform::Handler::upadteOemDbusPaths(
 void pldm::responder::oem_ibm_platform::Handler::_processSystemReboot(
     sdeventplus::source::EventBase& /*source */)
 {
+    BiosAttributeList biosAttrList;
+    biosAttrList.push_back(std::make_pair("pvm_boot_initiator", "Host"));
+    setBiosAttr(biosAttrList);
     pldm::utils::PropertyValue value =
         "xyz.openbmc_project.State.Chassis.Transition.Off";
     pldm::utils::DBusMapping dbusMapping{"/xyz/openbmc_project/state/chassis0",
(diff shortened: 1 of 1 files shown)


sampmisr (approved):
LGTM.will wait for Andrew's comments

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1188,6 +1188,9 @@ void pldm::responder::oem_ibm_platform::Handler::upadteOemDbusPaths(
 void pldm::responder::oem_ibm_platform::Handler::_processSystemReboot(
     sdeventplus::source::EventBase& /*source */)
 {
+    BiosAttributeList biosAttrList;
```
geissonator, line 1191: So just to double check, the only time we go into this _processSystemReboot() function is when we are going through the inband update path? Or really, we only go through this path when the host is responsible for asking for the BMC reboot and re-ipl after that?

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1423,7 +1426,8 @@ void pldm::responder::oem_ibm_platform::Handler::handleBootTypesAtPowerOn()
     BiosAttributeList biosAttrList;
     auto bootInitiator = getBiosAttrValue("pvm_boot_initiator_current");
     std::string restartCause;
-    if (bootInitiator != "HMC" && !bootInitiator.empty())
+    if (((bootInitiator != "HMC") || (bootInitiator != "Host")) &&
```
geissonator, line 1429: Did you get a chance to confirm that when we boot with "Host" that PHYP resets it once they are booted? We need to be sure "someone" clears it or it will never be cleared.

sagisin: yes, I did confirm that on Poweron PHYP resets these 2 bios attributes to default values.

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1477,32 +1481,7 @@ void pldm::responder::oem_ibm_platform::Handler::handleBootTypesAtChassisOff()
             << "ERROR in fetching the pvm_boot_initiator and pvm_boot_type BIOS attribute values\n";
         return;
     }
-    else if ((bootInitiator == "Auto") && (bootType == "IPL"))
-    {
-        std::cerr << "Setting the APR to AlwaysOn as pvm_boot_initiator="
-                  << bootInitiator << " and pvm_boot_type=" << bootType
-                  << std::endl;
-#if 0
```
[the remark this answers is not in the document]
sagisin: Yeah we have introduced a new effecter with new state PLDM_POWER_OFF_HARD_GRACEFUL for PHYP to trigger, which will set the APR settings in that path. And that has been tested.

geissonator (approved):

PR #180, opened 2022-03-17
PLDM: Fix code update Rename

sagisin:
This commit is to fix the bootside rename
defect:  SW544814 in the out of band code
update path.


anoo1 (approved):

geissonator:
Still need @sampmisr and @manojkiraneda to take a look but in the interest of time, merge now so we can get through CI.

PR #183, opened 2022-03-22
PLDM should not honor FILE IO read/write with 0 byte

sagisin:


jaypadath (changes requested):
if (length % dma::minSize)

Line no 299 also have the check. I think we need to update this condition too ?

sagisin:
> if (length % dma::minSize)
> 
> Line no 299 also have the check. I think we need to update this condition too ?

I have added the changes

jaypadath (approved):

PR #184, opened 2022-03-25
Fix Race condition in PDR generation

manojkiraneda:
This PR would fix mutiple issues :
1.When ever pldm is started , it sends a getPLDMVersion
  command to host and if we get a response we set HostUp
  to true. But we started to see cases where PHYP does not
  respond to getPLDMVersion but instead does a GetPDR command.
  So we have a minimal window where we could go ahead and create
  PDR's even when BMC is not in Ready State.

2.Check for other BMC states and don't allow anyone to create PDR's
  if BMC is not in Ready state.


lxwinspur (approved):
Looks good to me

Pavithrab7 (approved):

sampmisr (approved):

PR #190, opened 2022-04-01
PLDM: Remove old boot side setting

sagisin:
This commit removes the old boot side setting of
bios attribute pvm_fw_boot_side which is now
deprecated.

Also adds an additional check before setting the
fw_boot_side bios attribute before setting the new
value which is set by host to make sure that the
bios attibute table is initiated before we try to
set the value.

### DIFF
--- a/libpldmresponder/bios_config.cpp
+++ b/libpldmresponder/bios_config.cpp
@@ -447,9 +447,13 @@ void BIOSConfig::updateBaseBIOSTableProperty()
                                           dbusProperties, "Set");
         std::variant<BaseBIOSTable> value = baseBIOSTableMaps;
 #ifdef OEM_IBM
+        const fs::path bootSideDirPath = "/var/lib/pldm/bootSide";
         for (const auto& [attrName, biostabObj] : baseBIOSTableMaps)
         {
-            if (attrName == "fw_boot_side")
+            // The additional check to see if /var/lib/pldm/bootSide file exists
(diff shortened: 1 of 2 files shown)


libpldmresponder/bios_config.cpp
```diff
@@ -447,9 +447,11 @@ void BIOSConfig::updateBaseBIOSTableProperty()
                                           dbusProperties, "Set");
         std::variant<BaseBIOSTable> value = baseBIOSTableMaps;
 #ifdef OEM_IBM
+        std::string dirPath = "/var/lib/pldm/bootSide";
+        const fs::path bootSideDirPath = dirPath;
```
manojkiraneda, line 451: we can use a single statment to do the fs path initialisation, first declaring string and then converting to fs:path object is redundant

libpldmresponder/bios_config.cpp
```diff
@@ -447,9 +447,11 @@ void BIOSConfig::updateBaseBIOSTableProperty()
...
         for (const auto& [attrName, biostabObj] : baseBIOSTableMaps)
         {
-            if (attrName == "fw_boot_side")
+            if ((attrName == "fw_boot_side") && fs::exists(bootSideDirPath))
```
Pavithrab7, line 454: Can you add a comment on why this check is needed here?

manojkiraneda (approved):
Looks good to me

PR #192, opened 2022-04-01
support a BootProgress of SystemSetup for host running

geissonator:
SystemSetup is a BootProgress that indicates the host is booted, but in
a state where user interaction is needed. On an IBM system, this state
is entered when VMI needs some configuration or when PHYP's NVRAM has
been cleared.

pldm needs to consider the host running in this scenario because it's
needed to send user configuration information like the VMI IP and
static vs. dhcp information.


anoo1 (approved):
These are the only files where BootProgress is checked.

PR #193, opened 2022-04-04
PLDM: Minimise sending BIOS attr change event

sagisin:
This commits add code to minimise sending redundant bios
attribute change events to host.

Defect:SW546927


anoo1 (approved):
It makes sense to check the existing NextBootSide value, and do not send a notification if it's the same as the requested value.
At a higher level, pldm could look into enhancing its logic for setting bios attributes where if the requested value was the same as the existing value, it would not send a notification to PHYP.

manojkiraneda (approved):

rfrandse:
fixes SW546927

PR #195, opened 2022-04-06
Fix the DBus persist bug

lxwinspur:
There is a bug here: if bmc is started or restarted and the host is
closed, all attribute values of the core type should not be restored
at this time, because when the host is powered on again, all the PDRs
of the core type will be sent again, so there will be bmc
Inconsistent with the state of the host's cores.

Signed-off-by: George Liu <liuxiwei@inspur.com>


lxwinspur:
link : https://github.com/ibm-openbmc/pldm/pull/201

PR #203, opened 2022-04-19
Remove Default Option

zamiseck:
Default should not be an option for the current eBMC system. This
option was applicable for FSP and ASMI based systems. Rainier and
Everest based systems should not have this option. The default
choice needs to be removed from pvm_default_os_type and
pvm_default_os_type_current, and the default value needs to be
changed to AIX.

Defect SW547992 :
https://w3.rchland.ibm.com/projects/SLIC/bestquest/index.php?defect=SW547992

### DIFF
--- a/oem/ibm/configurations/bios/enum_attrs.json
+++ b/oem/ibm/configurations/bios/enum_attrs.json
@@ -194,11 +194,10 @@
          "possible_values":[
             "AIX",
             "Linux",
-            "IBM I",
-            "Default"
+            "IBM I"
          ],
          "default_values":[
-            "Default"
+            "AIX"
          ],
          "helpText" : "CEC Primary OS",
          "displayName" : "CEC Primary OS"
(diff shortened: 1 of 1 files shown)


manojkiraneda:
I am not sure , the intention and the reason behind the fix  - the Default value was added because of a manufacturing requirement - and phyp has enough logic to decided which os it need to boot (if the bios attribute is set to Default)

mzipse:
@manojkiraneda why does MFG need "Default"?  And what is the behavior that is expected.  There is nothing documented in the BIOS spec.  The GUI is picking the choices up directly from Redfish BIOS attribute and displaying this "Default" value as a choice for customers to pick from but noone knows what it's for or when would a customer use it.

Assuming we keep it, we need to get it added to the spec.

manojkiraneda:
The Default option was added to let phyp read the sys_rb keyword and decide which operating system to boot. the default a way to tell that no one has set a specific operating system and let phyp decide which is to boot.

manojkiraneda:
Agree we need to add it to the spec

manojkiraneda:
https://ibm-systems-power.slack.com/archives/CD62M0R46/p1650903356658309

More discussion on this ^^

PR #204, opened 2022-04-19
oem: ibm: Close fd after transferFileData

anoo1:
The file descriptor was being closed before the call to transferFileData
which performs a write to it, move the close call after the call to
write. Also add an fsync call to ensure the data has made it to the
filesystem and can be processed.

Tested: Performed an inband update.

Change-Id: I199782e2a1b96496d57f3f6e840a298ba66de9c5

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_lid.hpp
+++ b/oem/ibm/libpldmresponder/file_io_type_lid.hpp
@@ -152,14 +152,15 @@ class LidHandler : public FileHandler
                       << "\n";
             return PLDM_ERROR;
         }
-        close(fd);
-
         rc = transferFileData(lidPath, false, offset, length, address);
         if (rc != PLDM_SUCCESS)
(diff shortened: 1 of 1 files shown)


oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -152,14 +152,15 @@ class LidHandler : public FileHandler
                       << "\n";
             return PLDM_ERROR;
         }
-        close(fd);
-
         rc = transferFileData(lidPath, false, offset, length, address);
```
anoo1, line 155: I just noticed that this transferFileData() call doesn't use the file descriptor, so I guess the file descriptor can be removed?

anoo1: Seems this may be for creating the file.

PR #205, opened 2022-04-20
PLDM: PCIE topology support

Pavithrab7:
This commit adds support to implement the topology interface
defined here - https://github.com/ibm-openbmc/phosphor-dbus-interfaces/blob/1020/yaml/com/ibm/PLDM/PCIeTopology.interface.yaml
and also host it under PLDM and also supports the set and get
property under the topology interface.

When PCIeTopologyRefresh property is set to true, we send the
setEffecterCall to host to set the effecter state 1 and 2.
When SaveTopologyRefresh property is set to true we send the
setEffecterCall to host to set the effecter state to 3 and then
sets it back to false.

Tested using busctl commands.

### DIFF
--- a/host-bmc/dbus/pcie_topology.cpp
+++ b/host-bmc/dbus/pcie_topology.cpp
@@ -1,0 +1,175 @@
+#include "pcie_topology.hpp"
+
+#include "libpldm/entity.h"
+#include "oem/ibm/libpldm/state_set_oem_ibm.h"
+
+#include "host-bmc/dbus/custom_dbus.hpp"
+
+namespace pldm
+{
+namespace dbus
+{
+
+bool PCIETopology::pcIeTopologyRefresh() const
+{
+    return sdbusplus::com::ibm::PLDM::server::PCIeTopology::
+        pcIeTopologyRefresh();
+}
+
+bool PCIETopology::updateTopologyRefresh()
+{
+    stateOfCallback.first = false;
+    stateOfCallback.second = false;
+    return sdbusplus::com::ibm::PLDM::server::PCIeTopology::pcIeTopologyRefresh(
+        true);
+}
+
+bool PCIETopology::callbackGetPCIeTopology(bool value)
+{
+    stateOfCallback.first = value;
+
+    if (stateOfCallback.first && stateOfCallback.second)
+    {
+        // if both the states of the effecter are set only then update the
+        // property
+        return updateTopologyRefresh();
+    }
+    return false;
+}
+
+bool PCIETopology::callbackGetCableInfo(bool value)
+{
+    stateOfCallback.second = value;
+
+    if (stateOfCallback.first && stateOfCallback.second)
+    {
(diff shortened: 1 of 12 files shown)


host-bmc/dbus/pcie_topology.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+    else
+    {
+        stateField.push_back({PLDM_REQUEST_SET, GET_PCIE_TOPOLOGY});
+        stateField.push_back({PLDM_REQUEST_SET, GET_CABLE_INFO});
+    }
+
+    if (value)
+    {
+        if (hostEffecterParser)
```
manojkiraneda, line 44: nit : both if's can me merged with &&

host-bmc/dbus/pcie_topology.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+    else
+    {
+        stateField.push_back({PLDM_REQUEST_SET, SAVE_PCIE_TOPLOGY});
+    }
+
+    if (value)
+    {
+        if (hostEffecterParser)
+        {
```
manojkiraneda, line 109: same here

host-bmc/dbus/pcie_topology.cpp
```diff
@@ -0,0 +1,144 @@
...
+                auto pdr =
+                    reinterpret_cast<pldm_state_effecter_pdr*>(rep.data());
+                effecterID = pdr->effecter_id;
+                break;
+            }
+
+            hostEffecterParser->sendSetStateEffecterStates(
+                mctpEid, effecterID, 1, stateField, nullptr, value);
+            return sdbusplus::com::ibm::PLDM::server::PCIeTopology::
+                savePCIeTopologyInfo(false);
```
manojkiraneda, line 136: all of these seems like common code for both the properties , can we make a function for these ? and call it with parameters ?

host-bmc/dbus/pcie_topology.cpp
Pavithrab7: done

manojkiraneda (approved):

host-bmc/dbus/pcie_topology.cpp
```diff
@@ -0,0 +1,130 @@
...
+        if (effecterID == 0)
+        {
+            return false;
+        }
+        hostEffecterParser->sendSetStateEffecterStates(
+            mctpEid, effecterID, 1, stateField,
+            std::bind(
+                std::mem_fn(&pldm::dbus::PCIETopology::updateTopologyRefresh),
+                this, std::placeholders::_1),
+            value);
```
manojkiraneda, line 55: ah , pavitra  i missed this  - you need to set the effecters with both 1 and 2

Pavithrab7: done

sampmisr (approved):

PR #217, opened 2022-04-27
PLDM: Issue ForceWarmReboot instead if Reboot

sagisin:
When PHYP toggles the PLDM_POWER_CYCLE_OFF_SOFT_GRACEFUL
we set the state manager property RequestedHostTransition
to ForceWarmReboot instead of Reboot, this will directly
reboot the host firmware without notification and chassis
power will remain on throughout the reboot, instead of
doing a graceful reboot

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
@@ -1672,7 +1672,7 @@ void pldm::responder::oem_ibm_platform::Handler::
     processPowerCycleOffSoftGraceful()
 {
     pldm::utils::PropertyValue value =
-        "xyz.openbmc_project.State.Host.Transition.Reboot";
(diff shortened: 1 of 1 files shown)


oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1672,7 +1672,7 @@ void pldm::responder::oem_ibm_platform::Handler::
     processPowerCycleOffSoftGraceful()
 {
     pldm::utils::PropertyValue value =
-        "xyz.openbmc_project.State.Host.Transition.Reboot";
+        "xyz.openbmc_project.State.Host.Transition.ForceWarmReboot";
```
manojkiraneda, line 1675: although this seems like a wierd combination where we do forcewarmreboot - when the state set in pldm is soft off graceful , but approving is as phyp team wants this behavior.

rfrandse:
fixes SW549067

PR #219, opened 2022-04-28
Utility to get object Path using location and inventory type

ArchanaKakani:
This function is added to support PCIe-Topology.
Util function takes Location code and Inventory Item Type
as inputs and return the dbus object path.

Tested:
  Reset Reload operation and normal poweon workrs fine.
  Also executed utility function with valid/invalid sets of input
  and got expected results.

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -390,6 +390,19 @@ PropertyValue DBusHandler::getDbusPropertyVariant(
     return value;
 }
 
+ObjectValueTree DBusHandler::getManagedObj(const char* service,
+                                           const char* rootPath)
+{
+    ObjectValueTree objects;
+    auto& bus = DBusHandler::getBus();
+    auto method = bus.new_method_call(service, rootPath,
+                                      "org.freedesktop.DBus.ObjectManager",
+                                      "GetManagedObjects");
+    auto reply = bus.call(method);
+    reply.read(objects);
+    return objects;
+}
+
 PropertyValue jsonEntryToDbusVal(std::string_view type,
                                  const nlohmann::json& value)
 {
--- a/libpldmresponder/fru.cpp
+++ b/libpldmresponder/fru.cpp
@@ -157,17 +157,12 @@ void FruImpl::buildFRUTable()
                          panelHotplugMatch);
 
     fru_parser::DBusLookupInfo dbusInfo;
-    // Read the all the inventory D-Bus objects
-    auto& bus = pldm::utils::DBusHandler::getBus();
 
+    // Read the all the inventory D-Bus objects
     try
     {
         dbusInfo = parser.inventoryLookup();
-        auto method = bus.new_method_call(
-            std::get<0>(dbusInfo).c_str(), std::get<1>(dbusInfo).c_str(),
-            "org.freedesktop.DBus.ObjectManager", "GetManagedObjects");
-        auto reply = bus.call(method);
-        reply.read(objects);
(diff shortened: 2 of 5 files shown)


common/utils.cpp
```diff
@@ -685,5 +706,37 @@ std::string getCurrentSystemTime()
     return ss.str();
 }
 
+std::string getLocationObject(std::string locationCode,
```
manojkiraneda, line 709: this cannot be in generic code ..move this to oem.

This is a very specific use case for ibm (i dont think any one would have this kind of requirement)

common/utils.cpp
```diff
@@ -390,6 +390,27 @@ PropertyValue DBusHandler::getDbusPropertyVariant(
...
+                                          "GetManagedObjects");
+        auto reply = bus.call(method);
+        reply.read(objects);
+    }
+    catch (const std::exception& e)
+    {
+        std::cerr << "Look up of objects failed : " <<  service << std::endl;
+    }
+
+    return objects;
```
manojkiraneda, line 411: is the dbus call fails ..what should be returned ?

a util function should not really have exception handling - its the function that calls this API that needs exception handling

common/utils.cpp
ArchanaKakani: Moved this function to oem/ibm

common/utils.cpp
ArchanaKakani: Handled the exception in caller function

libpldmresponder/fru.cpp
```diff
@@ -157,17 +157,12 @@ void FruImpl::buildFRUTable()
                          panelHotplugMatch);
 
     fru_parser::DBusLookupInfo dbusInfo;
-    // Read the all the inventory D-Bus objects
-    auto& bus = pldm::utils::DBusHandler::getBus();
+    dbusInfo = parser.inventoryLookup();
```
manojkiraneda, line 160: why was this moved out of try ?

oem/ibm/libpldmresponder/utils.cpp
```diff
@@ -501,6 +507,48 @@ void hostPCIETopologyIntf(
         "/xyz/openbmc_project/pldm", mctp_eid, hostEffecterParser);
 }
 
+std::string getLocationObject(std::string locationCode,
+                              std::string inventoryItemType)
```
manojkiraneda, line 511: std::string getObjectPathByLocationCode(const std::string& locationcode, const std::string& inventoryItemType)

oem/ibm/libpldmresponder/utils.cpp
```diff
@@ -501,6 +507,48 @@ void hostPCIETopologyIntf(
...
+{
+    std::string path("");
```
manojkiraneda, line 513: the default constructor of std::string -> would construct an empty string.

libpldmresponder/fru.cpp
ArchanaKakani: Moved the function inside try block

oem/ibm/libpldmresponder/utils.cpp
ArchanaKakani: Done

oem/ibm/libpldmresponder/utils.cpp
ArchanaKakani: Done

manojkiraneda (approved):

sampmisr (approved):

PR #222, opened 2022-04-29
oem: ibm: Do not remove the staging directory

anoo1:
The staging directory is created by the software manager application,
therefore it should not be deleted by pldm or it won't be re-created
until the next BMC reboot, and the inband update would fail with:

Apr 29 20:21:53 p10bmc pldmd[1980]: Cannot stat source directory
"/var/lib/phosphor-software-manager/hostfw/staging/lid" because No such
file or directory

Instead iterate over the contents of the directory to clear them.

Tested: The contents of the staging directory got removed, but not the
staging directory itself.

Change-Id: Id76ac325d49ca48891c99e7b71b1a0d5d64c85c3

### DIFF
--- a/oem/ibm/libpldmresponder/inband_code_update.cpp
+++ b/oem/ibm/libpldmresponder/inband_code_update.cpp
@@ -516,12 +516,13 @@ void CodeUpdate::setOemPlatformHandler(
 
 void CodeUpdate::clearDirPath(const std::string& dirPath)
 {
-    if (!fs::exists(dirPath))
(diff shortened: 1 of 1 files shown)


manojkiraneda (approved):
I had made this change earlier , when i saw pldm crashing while it is trying to iterate over a directory which is not present. So used the remove_all and i did not realize the consequence.

Seems like you have fixed it using is_directory(), looks good to me. Thanks Adriana.

PR #223, opened 2022-04-29
oem: ibm: Add squashfs options for reproducible builds

anoo1:
The mksquashfs command adds data to the header such as the creation
time, this makes the generated squashfs image not match the one
generated by the build. Use additional options offered by the squashfs
tools for reproducible builds such as setting the timestamp to 0
(reference:
https://lore.kernel.org/lkml/CAB3wodcL=gnQOmHGGNukWK3OUbU2p=OHzLmzPi7ns_WNTGBEwg@mail.gmail.com/)

Also specify the permissions (777 for the directory to allow any BMC
user name to access this directory to write the lids), and 440 for the
lid files as these files are intended to be read-only, for cases the
root filesystem on the BMC is not read-only.

Tested: Verified the generated image was identical to one generated
during a build process.

Change-Id: Id3aeebafad2395bbf715a701c8592cd6939f57bf

### DIFF
--- a/oem/ibm/libpldmresponder/inband_code_update.cpp
+++ b/oem/ibm/libpldmresponder/inband_code_update.cpp
@@ -671,8 +671,25 @@ int processCodeUpdateLid(const std::string& filePath)
         return PLDM_ERROR;
     }
 
-    fs::create_directories(imageDirPath);
-    fs::create_directories(lidDirPath);
+    if (!fs::exists(imageDirPath))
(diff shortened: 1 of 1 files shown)


manojkiraneda (approved):

anoo1:
> End-to-end testing scheduled for next week. Will post the results here.

Testing complete, the image is now identical to the one generated by a build process.

PR #230, opened 2022-05-08
Improve serialization performance in pldm

manojkiraneda:
- Observed that json serialization archive format in cereal
  library is very slow with the current pldm infrastructure
  and its blocking pldm for almost 40 seconds(for a system that
  has 128 cores and 4 mex drawers connected) after every PDR
  exchange.
- When pldm is blocked , it cannot respond to messages from other
  remote entities.

- Moving from json serialisation archive format to binary serialization
  format in cereal has made things alot faster. I see that now the
  serailization is much faster and just took around 1-2 seconds to
  complete(for a system that has 128 core dbusobjects & 224 mex dbus objects).

- This commit also adds support for custom store, custom restore functionality
  where pldm can serialize and deserialize the dbus objects based on a
  configuration file.

- This commit also add necessary checks to not serialise the object again
  and again if the object value does not change.

Tested By :

- Multiple power on operations on a max configured everest machine & never
  saw a reset reload.


lxwinspur (approved):

lxwinspur (approved):

sampmisr (approved):

manojkiraneda:
fixes : SW549012

PR #231, opened 2022-05-08
Fix pldm crash caused by corrupted iterator

manojkiraneda:
- When PDR exchange is completed with phyp, pldm will Query
  HBRT and PHYP for the state of the host sensors in a recursive
  loop. and in the middle of that recursion if some one power off the
  system, the iterator is being corrupted and any access further using
  the iterator would crash pldm.

TestedBy:

- Used automation test on the pldm change and it went successful without
any problem for  5 complete iterations of the test.


sampmisr (approved):

manojkiraneda:
fixes : SW550060

PR #232, opened 2022-05-09
[PLDM FRU]: Removing FRU Record from FRU Record Table

ArchanaKakani:
When a CEC FRU such as a Fan or Power Supply is concurrently removed,
BMC properly sends event to remove the FRU Record Set ID and to
 modify the Entity Association PDRs.
BMC also removes the corresponding state sensors/ effecters PDRs.
However, BMC does not remove the FRU entry from the FRU Record Table

Modified code to delete FRU Record from FRU Record table.

Tested: Using busctl command and pldmtool
 busctl set-property xyz.openbmc_project.Inventory.Manager
      /xyz/openbmc_project/inventory/system/chassis/motherboard/fan0
     xyz.openbmc_project.Inventory.Item Present b false

 pldmtool fru GetFruRecordTable -v

### DIFF
--- a/libpldmresponder/fru.cpp
+++ b/libpldmresponder/fru.cpp
@@ -393,9 +393,7 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
      pldm_entity_association_tree_visit(entityTree,&out, &num);
      free(out);*/
     // sm00
-
     pldm_entity_association_tree_delete_node(entityTree, removeEntity);
-
     // sm00
     /*std::cout << "\nprinting the entityTree after deleting node\n";
     num = 0;
@@ -414,23 +412,8 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
               << removeEntity.entity_container_id << " ]\n";
     associatedEntityMap.erase(fruObjPath); // sm00
 
-    /* if (table
-             .size()) /// need to remove the entry from table before doing this
-                      // may be a separate commit to handle that. as of now it
-                      // does not create issue because the pdr repo is updated.
-                      // the fru record table is not. which means the pldmtool
-                      // fru commands will still show the old record. that may
-     be
-                      // fine since Host is not asking at this moment they are
-                      // getting the updated pdr (both fru and entity assoc pdr)
-                      // but eventually we need it because an add happened after
-                      // a remove will cause two fru records for the same fan
-     {
-         padBytes = pldm::utils::getNumPadBytes(table.size());
-         table.resize(table.size() + padBytes, 0);
-         // Calculate the checksum
-         checksum = crc32(table.data(), table.size());
-     }*/
+    deleteFruRecord(rsi);
+
     sendPDRRepositoryChgEventbyPDRHandles(
         std::move(std::vector<ChangeEntry>(1, deleteRecordHdl)),
         std::move(std::vector<uint8_t>(1, PLDM_RECORDS_DELETED)));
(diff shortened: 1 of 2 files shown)


sagisin (approved):
looks good to me

libpldmresponder/fru.cpp
```diff
@@ -418,6 +418,7 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
               << removeEntity.entity_container_id << " ]\n";
     associatedEntityMap.erase(fruObjPath); // sm00
 
+    deleteFruRecord(rsi);
     /* if (table
```
jaypadath, line 422: Does this commented explanation makes any sense as we are deleting the fru record table entry now ?

libpldmresponder/fru.cpp
```diff
@@ -453,6 +454,43 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
...
+    const struct pldm_fru_record_tlv* tlv;
+    size_t pos = 0;
+
+    while (table.size() > pos)
+    {
+        size_t recordLen = sizeof(struct pldm_fru_record_data_format) -
+                           sizeof(struct pldm_fru_record_tlv);
+        tlv = recordSetSrc->tlvs;
+
+        for (int i = 0; i < recordSetSrc->num_fru_fields; i++)
```
jaypadath, line 473: Please use unsingned int (uint) instead of signed one

ArchanaKakani: Done

libpldmresponder/fru.cpp
ArchanaKakani: Removed the comment

libpldmresponder/fru.cpp
```diff
@@ -399,8 +399,6 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
      free(out);*/
     // sm00
 
-    pldm_entity_association_tree_delete_node(entityTree, removeEntity);
```
Pavithrab7, line 402: why is this removed?

libpldmresponder/fru.cpp
```diff
@@ -453,6 +436,43 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
...
+            size_t len = sizeof(*tlv) - 1 + tlv->length;
+            recordLen += len;
+            tlv = reinterpret_cast<const struct pldm_fru_record_tlv*>(
+                (char*)tlv + len);
+        }
+        if ((recordSetSrc->record_set_id != htole16(rsi) && rsi != 0))
+        {
+            std::copy(table.begin() + pos, table.begin() + pos + recordLen,
+                      std::back_inserter(updatedFruTbl));
+        }
```
manojkiraneda, line 466: we are copying everything from begin to the postiion , but what about the records that are after the position ?
if the fru to be removed is at 5th postion , we just have to skip 5 an copy the entire table, but looking at this i feel we are copying only till 5th location.

Am i missing something here ?

manojkiraneda (changes requested):

libpldmresponder/fru.cpp
ArchanaKakani: We r running this if block in the while loop till we reach end of the frutable. So it will copy all other FRU records except the rsi matching ones.

libpldmresponder/fru.cpp
```diff
@@ -453,6 +436,43 @@ void FruImpl::removeIndividualFRU(const std::string& fruObjPath)
...
+            table.data());
+
+    const struct pldm_fru_record_tlv* tlv;
+    size_t pos = 0;
+
+    while (table.size() > pos)
+    {
+        size_t recordLen = sizeof(struct pldm_fru_record_data_format) -
+                           sizeof(struct pldm_fru_record_tlv);
+        tlv = recordSetSrc->tlvs;
```
manojkiraneda, line 453: null check missing before de-reference.

manojkiraneda (approved):

libpldmresponder/fru.cpp
ArchanaKakani: Done

PR #234, opened 2022-05-10
[OEM_IBM]:Power off chassis in case of Hard Poweroff Event

ArchanaKakani:
In case of PLDM_POWER_OFF_HARD_GRACEFUL event PLDM shall update APR
and poweroff Chassis.

The OEM PowerDown effecter was defined to support 3 paths.
1. phyp initiated normal shutdown - PLDM_POWER_OFF_SOFT_GRACEFUL
2. phyp initiated re-ipl - PLDM_POWER_CYCLE_OFF_SOFT_GRACEFUL
3. phyp shutdown due to UPS event - PLDM_POWER_OFF_HARD_GRACEFUL

In all 3 cases, phyp has shut things down and is supposed
to be powered off.

When phyp receives the UPS event, phyp shuts things down
After phyp shuts down and sets the boot parms, phyp sets the OEM
PowerDown effecter to PLDM_POWER_OFF_HARD_GRACEFUL.
Then BMC should power off phyp.

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
@@ -1671,6 +1671,7 @@ void pldm::responder::oem_ibm_platform::Handler::updateContainerIDofProcLed()
 void pldm::responder::oem_ibm_platform::Handler::
     processPowerCycleOffSoftGraceful()
 {
+    std::cerr << "Received soft graceful power cycle request" << std::endl;
     pldm::utils::PropertyValue value =
         "xyz.openbmc_project.State.Host.Transition.ForceWarmReboot";
     pldm::utils::DBusMapping dbusMapping{"/xyz/openbmc_project/state/host0",
@@ -1690,6 +1691,7 @@ void pldm::responder::oem_ibm_platform::Handler::
 
 void pldm::responder::oem_ibm_platform::Handler::processPowerOffSoftGraceful()
 {
+    std::cerr << "Received soft power off graceful request" << std::endl;
     pldm::utils::PropertyValue value =
         "xyz.openbmc_project.State.Chassis.Transition.Off";
     pldm::utils::DBusMapping dbusMapping{"/xyz/openbmc_project/state/chassis0",
@@ -1709,6 +1711,7 @@ void pldm::responder::oem_ibm_platform::Handler::processPowerOffSoftGraceful()
 
 void pldm::responder::oem_ibm_platform::Handler::processPowerOffHardGraceful()
 {
+    std::cerr << "Received hard power off graceful request" << std::endl;
     pldm::utils::PropertyValue value =
         "xyz.openbmc_project.Control.Power.RestorePolicy.Policy.AlwaysOn";
     pldm::utils::DBusMapping dbusMapping{
@@ -1725,6 +1728,7 @@ void pldm::responder::oem_ibm_platform::Handler::processPowerOffHardGraceful()
             << "Setting one-time restore policy failed, Unable to set property PowerRestorePolicy. ERROR="
             << e.what() << "\n";
     }
+    processPowerOffSoftGraceful();
 }
 
 void pldm::responder::oem_ibm_platform::Handler::setSurvTimer(bool value)


oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1705,6 +1705,7 @@ void pldm::responder::oem_ibm_platform::Handler::processPowerOffSoftGraceful()
             << "Error in powering down the host. Unable to set property RequestedPowerTransition. ERROR="
             << e.what() << "\n";
     }
+    processPowerCycleOffSoftGraceful();
```
geissonator, line 1708: I'm confused. I thought the issue was that PHYP was calling processPowerOffHardGraceful() to set the APR, and in that function, we need to add the call to hard power off (processPowerOffSoftGraceful())

sagisin: so the expectation from PHYP was they when they call processPowerOffHardGraceful(), we need to set the APR and also chassis off. And what we were already doing is setting the APR. 
Here we have just added code to do the Chassisoff  after the APR, and as processPowerCycleOffSoftGraceful() already does that we have just reused the code here.

sagisin: Ah I got what you were trying to say. @ArchanaKakani could you please move the processPowerCycleOffSoftGraceful() inside processPowerOffHardGraceful() after setting the APR please.

sagisin (commented):
@ArchanaKakani could you please change the commit message as discussed?

manojkiraneda (approved):

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1725,6 +1728,7 @@ void pldm::responder::oem_ibm_platform::Handler::processPowerOffHardGraceful()
             << "Setting one-time restore policy failed, Unable to set property PowerRestorePolicy. ERROR="
             << e.what() << "\n";
     }
+    processPowerCycleOffSoftGraceful();
```
geissonator, line 1731: Still don't think this is right. It should be calling "processPowerOffSoftGraceful". We want to arm APR and then chassis power off the system. My state-manager code will call APR to power back on once UPS is fixed.

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1671,6 +1671,7 @@ void pldm::responder::oem_ibm_platform::Handler::updateContainerIDofProcLed()
 void pldm::responder::oem_ibm_platform::Handler::
     processPowerCycleOffSoftGraceful()
 {
+    std::cerr << "Received soft graceful power cycle request" << std::endl;
```
sagisin, line 1674: Please remove Debug trace

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1690,6 +1691,7 @@ void pldm::responder::oem_ibm_platform::Handler::
 
 void pldm::responder::oem_ibm_platform::Handler::processPowerOffSoftGraceful()
 {
+    std::cerr << "Received soft power off graceful request" << std::endl;
```
sagisin, line 1694: Please remove Debug trace

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1709,6 +1711,7 @@ void pldm::responder::oem_ibm_platform::Handler::processPowerOffSoftGraceful()
 
 void pldm::responder::oem_ibm_platform::Handler::processPowerOffHardGraceful()
 {
+    std::cerr << "Received hard power off graceful request" << std::endl;
```
sagisin, line 1714: Please remove Debug trace

sagisin (changes requested):

geissonator (commented):
ok looks good. I'm happy either way on the extra debug traces. They'll only be logged on power off's so I don't worry about the filling the journal or anything. One point of confusion is that the debug messages don't really align with the actual operation (because we've overridden them a bit for our purposes).

manojkiraneda:
> ok looks good. I'm happy either way on the extra debug traces. They'll only be logged on power off's so I don't worry about the filling the journal or anything. One point of confusion is that the debug messages don't really align with the actual operation (because we've overridden them a bit for our purposes).

Yes andrew, I agree that the traces don't really align with the dbus operations that we are doing , but we are trying to track phyp set effecter operation with those. and based on that we can try to deduce what is the bmc operation involved.

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
manojkiraneda: @sagisin i have requested archana to add traces ..as anything involving power offs e.t.c are critical operations and does not really happen much in field (so we dont really fill up the journal)

PR #241, opened 2022-05-17
PLDM: Reset image activation match

sagisin:
This commit adds code to reset the image activation signal
match every time after sending the end update sensor event
PHYP. This is done in order for pldm to receive the image
activation after every successful inband code update with
or without system reboot.

Test: Performed inband code update on IBMi and AIX OS.


oem/ibm/libpldmresponder/inband_code_update.cpp
```diff
@@ -387,6 +388,7 @@ void CodeUpdate::setVersions()
                                                     uint8_t(CodeUpdateState::
                                                                 START));
                                                 newImageId.clear();
+                                                imageActivationMatch.reset();
                                             }
                                         }
```
manojkiraneda, line 393: why cant we clear the match after if & else if ? instead of at two places ?

manojkiraneda (approved):

PR #244, opened 2022-05-19
Change LanesInUse to int64_t

manojkiraneda:
LanesInUse is changed from size_t to int64_t to cater for
unknown case, this PR would align pldm changes to the new PDI
change.


sampmisr (approved):

Pavithrab7 (approved):
Can you add a link to that PDI change for reference?

rfrandse:
failing to compile because it's dependent on the PDI change

PR #253, opened 2022-05-23
oem: ibm: Move the NVRAM files to the hostfw directory

anoo1:
Move the NVRAM files from /var/ on the BMC to the hostfw directory so
that they don't get erased during a BMC factory reset.

To help migrate existing systems, check if the files exist in /var/, and
if they do, move them to the hostfw destination.

Tested: Verified that the PHYP NVRAM files were moved if they already
existed in /var/, and that empty files were created in hostfw if they
didn't.


oem/ibm/service_files/pldm-create-phyp-nvram.service
```diff
@@ -1,10 +1,9 @@
 [Unit]
 Description=Create empty PHYP-NVRAM file
-ConditionPathExists=!/var/lib/pldm/PHYP-NVRAM
+ConditionPathExists=!/var/lib/phosphor-software-manager/hostfw/nvram/PHYP-NVRAM
 
 [Service]
-ExecStart=/bin/mkdir -p /var/lib/pldm/
-ExecStart=/bin/dd if=/dev/zero of=/var/lib/pldm/PHYP-NVRAM bs=1024 count=145408
+ExecStart=/bin/sh -c "if [ -f /var/lib/pldm/PHYP-NVRAM ]; then mv /var/lib/pldm/PHYP-NVRAM /var/lib/phosphor-software-manager/hostfw/nvram/PHYP-NVRAM; else dd if=/dev/zero of=/var/lib/phosphor-software-manager/hostfw/nvram/PHYP-NVRAM bs=16 count=1; fi"
```
manojkiraneda, line 6: why is the block size and count changed ?

manojkiraneda: phyp cannot boot with 16bytes nvram , the original size was 142MB if i am not wrong.

anoo1: Fixed. I had wrongly copied the change from the cksum service file.

manojkiraneda (approved):

PR #254, opened 2022-05-25
[OEM_IBM]:Create LED PDRs for unpopulated DIMMs and Procs

ArchanaKakani:
Those changes are required to toggle the LEDs for unpopulated
Procs and DIMMs

So, irrespective of whether a DIMM or ProcessorModule is present,
host will always send Memory Board & Physical Socket ,
    Memory Module(DIMM) under Memory Board entiity
    Processor Module(DCM) under Physical Socket entity
so from pldm side, we need to modify the above mentioned led PDR's
with the same entity type , instance number & container id for both
Memory Board entity & Physical Socket entity from hostboot.

Tested:
Verified the PLDM tree using busctl command
Also tested deconfiguring the core using busctl command


manojkiraneda:
this change is contained in https://github.com/ibm-openbmc/pldm/pull/260

PR #257, opened 2022-05-26
PLDM: FCO and DIMM D-bus Objects persistency

sagisin:
Changes to persist the CPU core count and DIMM
Operational status across reboots when host is powered
off.

Defect:SW550463

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -295,21 +295,45 @@ void DBusHandler::setDbusProperty(const DBusMapping& dBusMap,
         auto& bus = getBus();
         auto service =
             getService(dBusMap.objectPath.c_str(), dBusMap.interface.c_str());
-        auto method = bus.new_method_call(
-            service.c_str(), dBusMap.objectPath.c_str(), dbusProperties, "Set");
-        if (dBusMap.objectPath ==
-                "/xyz/openbmc_project/network/hypervisor/eth0/ipv4/addr0" ||
-            dBusMap.objectPath ==
-                "/xyz/openbmc_project/network/hypervisor/eth1/ipv4/addr0")
+        if (service == "xyz.openbmc_project.Inventory.Manager")
         {
-            std::cout << " ,service :" << service.c_str()
-                      << " , interface : " << dBusMap.interface.c_str()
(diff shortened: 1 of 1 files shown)


common/utils.cpp
```diff
@@ -295,6 +295,28 @@ void DBusHandler::setDbusProperty(const DBusMapping& dBusMap,
...
+            auto service1 = getService("/xyz/openbmc_project/inventory",
+                                       "xyz.openbmc_project.Inventory.Manager");
+            auto method = bus.new_method_call(
+                service1.c_str(), "/xyz/openbmc_project/inventory",
+                "xyz.openbmc_project.Inventory.Manager", "Notify");
+            method.append(std::move(objectValueTree));
+            bus.call_noreply(method);
+        }
         auto method = bus.new_method_call(
             service.c_str(), dBusMap.objectPath.c_str(), dbusProperties, "Set");
```
manojkiraneda, line 321: even if the service is inventory manager why we are setting the property again ?

sagisin: Done

common/utils.cpp
```diff
@@ -295,21 +295,47 @@ void DBusHandler::setDbusProperty(const DBusMapping& dBusMap,
...
+                                std::get<0>(variant));
+            std::string objPath = dBusMap.objectPath.c_str();
+            std::string toReplace("/xyz/openbmc_project/inventory/system");
+            size_t pos = objPath.find(toReplace);
+            objPath.replace(pos, toReplace.length(), "/system");
+            interfaceMap.emplace(dBusMap.interface.c_str(), propertyMap);
+            objectValueTree.emplace(std::move(objPath),
+                                    std::move(interfaceMap));
+            auto service1 = getService("/xyz/openbmc_project/inventory",
+                                       "xyz.openbmc_project.Inventory.Manager");
```
manojkiraneda, line 313: this does not make sense, you already know the service name(at line number 298) when you entered the if statement, there is no need for us to do another mapper call to get the service name here.

manojkiraneda (approved):

jaypadath (approved):

PR #259, opened 2022-05-26
PLDM: Reset image activation after sending end update event

sagisin:
ImageActivation match which is reset even before we receive
the Image Active signal is received. Which is why we are
failing to send the End update sensor event to PHYP. This
needs to be reset only after the end update success/failure
event to PHYP after receiving the Active signal on the activation
property.

Defect:SW551662

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
@@ -146,6 +146,7 @@ int pldm::responder::oem_ibm_platform::Handler::
                 if (stateField[currState].effecter_state ==
                     uint8_t(CodeUpdateState::START))
                 {
+                    std::cout << "Received Start Update Request From PHYP\n";
                     codeUpdate->setCodeUpdateProgress(true);
                     startUpdateEvent =
                         std::make_unique<sdeventplus::source::Defer>(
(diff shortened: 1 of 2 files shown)


SrideviRamesh (approved):

manojkiraneda (approved):

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1198,6 +1203,7 @@ void pldm::responder::oem_ibm_platform::Handler::_processSystemReboot(
                                          "RequestedPowerTransition", "string"};
     try
     {
+        std::cout << "ChassifOff the host\n";
```
manojkiraneda, line 1206: can you add context in this ? like code update ?

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1234,6 +1240,7 @@ void pldm::responder::oem_ibm_platform::Handler::_processSystemReboot(
                             "Policy.AlwaysOn";
                     try
                     {
+                        std::cout << "Setting the one time APR policy\n";
```
manojkiraneda, line 1243: same here as well

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1249,6 +1256,7 @@ void pldm::responder::oem_ibm_platform::Handler::_processSystemReboot(
                     value = "xyz.openbmc_project.State.BMC.Transition.Reboot";
                     try
                     {
+                        std::cout << "Rebooting the BMC\n";
```
manojkiraneda, line 1259: same here as well

PR #262, opened 2022-05-26
PLDM: Disable surveillance timer

sagisin:
This commit is to disable the timer when we fail
to receive the surveillance pings from PHYP.

DEFECT:SW551173


jaypadath (approved):

jaypadath (approved):
Looks good to me

PR #269, opened 2022-06-02
PLDM Topology : PCIe Link Reset support

Pavithrab7:
This commit implements the link reset feature support
for both primary and secondary links.
When the link reset property is set to true, a host
stste effecter is triggered.

Tested using busctl commands.

### DIFF
--- a/host-bmc/dbus/linkreset.cpp
+++ b/host-bmc/dbus/linkreset.cpp
@@ -1,17 +1,91 @@
 #include "linkreset.hpp"
 
+#include "libpldm/entity.h"
+#include "libpldm/state_set.h"
+#include "libpldm/states.h"
+
 namespace pldm
 {
 namespace dbus
 {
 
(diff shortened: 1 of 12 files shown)


host-bmc/dbus/linkreset.cpp
```diff
@@ -1,17 +1,89 @@
...
+    if (stateEffecterPDRs.empty())
+    {
+        std::cerr
+            << "PCIe LinkReset: The state set PDR can not be found, entityType = "
+            << entityType << std::endl;
+        return effecterID;
+    }
+
+    uint32_t entityInstance =
+        pldm::responder::utils::getLinkResetInstanceNumber(path);
```
manojkiraneda, line 70: this should be under oem-ibm.

Pavithrab7: done

PR #275, opened 2022-06-08
Release the oldest instance id when all the instance ids are exhausted

manojkiraneda:
With this commit PLDM will :
1. Keep the timestamps for every instance id.
2. When all the instance ids are exhuasted , then pldm will
   look at the oldest instance id and release it.
3. pldm will only crash if none of the instance id's are older
   than the instance_id_expiration time.

### DIFF
--- a/pldmd/instance_id.cpp
+++ b/pldmd/instance_id.cpp
@@ -1,3 +1,5 @@
+#include "config.h"
+
 #include "instance_id.hpp"
 
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


pldmd/instance_id.hpp
```diff
@@ -18,16 +24,25 @@ class InstanceId
      */
     uint8_t next();
 
+    /** @brief Get the oldest instance id baed on timestamp
```
sagisin, line 27: Based?

pldmd/instance_id.cpp
```diff
@@ -15,10 +17,60 @@ uint8_t InstanceId::next()
...
+        auto instance = returnOldestId();
+        if (instance.has_value())
+        {
+            idx = instance.value();
+            // forcefully release the instance id
+            markFree(idx);
+        }
+        else
+        {
+            throw std::runtime_error("No free instance ids");
```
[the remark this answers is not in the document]
manojkiraneda: we will end up here, when none of the instance id's have a time stamp greater than INSTANCE_ID_EXPIRATION_INTERVAL.

That means all 32 instance id's are acquired just recently in span of less than 5 seconds - in  that case those instance ids can still be in use - in that case we will throw an exception and crash like usual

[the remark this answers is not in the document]
sagisin: got you, thanks.

sagisin (approved):

pldmd/instance_id.cpp
nmuruli, line 33: Can we say it as "Oldest Instance Id could not be found" ?  as Line 22 ia already saying no free ids

manojkiraneda: Done.

Pavithrab7 (approved):

PR #286, opened 2022-06-21
oem-ibm dump offload :Avoid Unmap of DMA on EINTR error  

raviteja-b:
These changes will handle the case where host crashes in the
middle of outofband DMA transfer during system dump offload.

When the host crashes during the DMA transfer,
PLDM receives EINTR(interrupt) and it unmaps the memory.
XDMA driver doesn't know that the operation is interrupted,
it still thinks that operation is in progress.
munmap will hang if xdma driver thinks the operation is in progress.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io.cpp
+++ b/oem/ibm/libpldmresponder/file_io.cpp
@@ -97,7 +97,16 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
         std::cerr
             << "transferHostDataToSocket : Failed to execute the DMA operation, RC="
             << rc << " ADDRESS=" << address << " LENGTH=" << length << "\n";
-        munmap(vgaMemDump, pageAlignedLength);
+        if (rc != -EINTR)
+        {
+            munmap(vgaMemDump, pageAlignedLength);
+        }
+        else
+        {
+            std::cerr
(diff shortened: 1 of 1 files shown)


sunharis (approved):

raviteja-b:
Fixes  https://w3.rchland.ibm.com/projects/bestquest/?defect=SW553074

oem/ibm/libpldmresponder/file_io.cpp
```diff
@@ -97,7 +97,16 @@ int DMA::transferHostDataToSocket(int fd, uint32_t length, uint64_t address)
...
-        munmap(vgaMemDump, pageAlignedLength);
+        if (rc != -EINTR)
+        {
+            munmap(vgaMemDump, pageAlignedLength);
+        }
+        else
+        {
+            std::cerr
+                << "Received interrupt during dump DMA transfer. Skipping Unmap"
+                << std::endl;
```
manojkiraneda, line 108: can you add a function name ? as well so that we know this came from what place ?

manojkiraneda: apart from this the changes look good.

raviteja-b: done

PR #290, opened 2022-06-28
Add a new service the clear's the stray DMA transactions

manojkiraneda:
In the current state , the DMA infrastruture written in pldm
is blocking IO, due to pldm operating in blocking mode there
are cases where pldm hangs in read/write system calls due to
multiple reasons like :

1. Stray DMA message in mctp buffer
When DMA operations in progress with PHYP and it goes down
abruptly, the DMA command of PHYP could be stored inside
the mctp buffer and that can be delivered to Hostboot(which does
not have pcie support) leading to a pldm hang.

2. PHYP crashes while a DMA transaction is in progress
xdma driver on bmc does not have a way to know if host had gone
down due to crash , so the outstanding DMA operation that is
currently running would be left hanging, leaving pldm to hang
as well.

Added new service pldm-clear-dma.service which would send
signals (SIGSTOP, SIGCONT) signals to pldm while the host
is going down, so that the stray DMA operations are cancelled
and pldm daemon can be recovered from the system call hangs.

### DIFF
--- a/oem/ibm/service_files/meson.build
+++ b/oem/ibm/service_files/meson.build
@@ -12,4 +12,11 @@ configure_file(
   install_dir: systemd_system_unit_dir,
   output: 'pldm-create-phyp-nvram-cksum.service',
 )
+configure_file(
+  copy: true,
+  input: 'pldm-clear-dma.service',
+  install: true,
+  install_dir: systemd_system_unit_dir,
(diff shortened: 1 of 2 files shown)


geissonator (approved):

geissonator:
Critical fix for SW553469. Must-fix approval in process.

PR #292, opened 2022-07-05
ensure proper dependency in phyp nvram services

geissonator:
Some changes were recently introduced that affected the startup order of
services on the way to bmc ready state. This change brought to light a
missing dependency in these service files. They depend on the nvram
links being setup correctly in var/lib/phosphor-software-manager/hostfw/
so therefore need to be sure the service that creates that link has run.

The change that added this dependency is not in upstream so this fix is
in downstream only.


manojkiraneda (approved):

manojkiraneda:
Looks goof to me, but would like a review from Adriana as well, as she is the last one touching this file.

PR #294, opened 2022-07-11
Don't discard the nvme drive presence

manojkiraneda:
There was a time where we were not able to find the
presence of the nvme drives on IBM systems,to work
around that pldm was looking at the parent of the drive
which is the nvme slot to figure out the presence of the
drive.

Now that we had code in platform-fru-detect which can
detect the presence of nvme drives over nvme-MI, the
workaround that we put to look for drives parent presence
is no longer needed. This PR is an attempt to fix that.


Pavithrab7 (approved):

manojkiraneda:
Fixes : SW554135

sampmisr (approved):

PR #311, opened 2022-08-26
Add support for previous state in the event messages

manojkiraneda:
In the current state, any sensor event from the bmc pldm
stack would have the same value for both the previous state
and the event state. This commit would fix that behaviour
by caching the previous state.

This commit would also add the previous state support for the
state sensor reading reply as well.

Tested By :
1. Change the value of sensor using busctl for example dimm8 identify
to false
Tx: 81 02 0a 01 00 00 4d 00 01 00 [ 01 00 ]

Note that the very first event for every sensor would have the previous
state as PLDM_SENSOR_UNKNOWN [0]

In the above example : [ event state : 01 , previous state : 00 ]

2. Change the value of dimm8 identify to true
Tx: 81 02 0a 01 00 00 4d 00 01 00 [ 02 01 ]

[ event state : 02 , previous state : 01 ]

3. Change the value of dimm8 identify to false
Tx: 81 02 0a 01 00 00 4d 00 01 00 [ 01 02 ]
Event state : 01 , previous state: 02

Change-Id: Ie4f452e1d27029ccc3846eb6bd882d48dda5d34c

### DIFF
--- a/host-bmc/dbus_to_event_handler.cpp
+++ b/host-bmc/dbus_to_event_handler.cpp
@@ -6,7 +6,6 @@
 
 namespace pldm
 {
-
 using namespace pldm::dbus_api;
 using namespace pldm::responder;
 using namespace pldm::responder::pdr;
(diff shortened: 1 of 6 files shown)


manojkiraneda:
jenkins run tests please

Pavithrab7 (approved):

PR #321, opened 2022-10-20
PLDM: Host lamp test asserted property not getting reset after lamp test finished.

Patel-Kamalkumar:
This commit contain:
- lamptest should complete in 4 min and after that asserted property of host and
  LED manager will be reset to false. but only LED manager asserted property
  was getting reset to false but host asserted property was remain true which is fixed
  with this fix.

- Fix of defect SW557624

### DIFF
--- a/oem/ibm/host-bmc/host_lamp_test.cpp
+++ b/oem/ibm/host-bmc/host_lamp_test.cpp
@@ -70,8 +70,7 @@ bool HostLampTest::asserted(bool value)
                 "Get Effecter ID failed, effecter = 0");
         }
     }
-
-    return value;
+    return sdbusplus::xyz::openbmc_project::Led::server::Group::asserted(value);
 }
 
 uint16_t HostLampTest::getEffecterID()


manojkiraneda (approved):

oem/ibm/host-bmc/host_lamp_test.cpp
```diff
@@ -70,6 +70,11 @@ bool HostLampTest::asserted(bool value)
                 "Get Effecter ID failed, effecter = 0");
         }
     }
+    else
```
Pavithrab7, line 73: instead of a new else block here can we not set the DBus property to false in line 37 itself?

Patel-Kamalkumar: line no 37 condition only hit if value=false and assertproperty=false.
but in lamp test assertproperty=true and we are trying to set value =false which will come to new Else block and make assert property to false.

ArchanaKakani: We enter line37 only if we are trying to set false value to the asserted property which is already false. In this case existing asserted property is true.

Pavithrab7 (approved):

oem/ibm/host-bmc/host_lamp_test.cpp
```diff
@@ -70,6 +70,11 @@ bool HostLampTest::asserted(bool value)
...
+    {
```
ArchanaKakani, line 74: instead of new else block we can set Asserted property at line 74 itself

ArchanaKakani (approved):

oem/ibm/host-bmc/host_lamp_test.cpp
Patel-Kamalkumar: modified as per comment.

jaypadath (approved):

PR #322, opened 2022-10-20
Removing topology and cable files before refresh

ArchanaKakani:
Topology/cables data received from host is written in a file by BMC. Data received after refresh is appended to the same file causes stale data fetched during DBus Object update.

Removing the topology file and cable attributes file just before receiving refreshed data so that latest data gets saved in the file.

Fixes: PE00DPCM

Tested:
Fetching/Refreshing Topology data:
  All the links/data were populated correctly.
  Topology and cable files were deleted and created on every refresh.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_pcie.cpp
+++ b/oem/ibm/libpldmresponder/file_io_type_pcie.cpp
@@ -47,6 +47,7 @@ std::unordered_map<linkId_t, linkType_t> PCIeInfoHandler::linkTypeInfo;
 PCIeInfoHandler::PCIeInfoHandler(uint32_t fileHandle, uint16_t fileType) :
     FileHandler(fileHandle), infoType(fileType)
 {
+    deleteTopologyFiles();
     receivedFiles.emplace(infoType, false);
 }
 int PCIeInfoHandler::writeFromMemory(
@@ -1103,6 +1104,25 @@ int PCIeInfoHandler::newFileAvailableWithMetaData(uint64_t /*length*/,
                                                   uint32_t /*metaDataValue4*/)
 {
     return PLDM_ERROR_UNSUPPORTED_PLDM_CMD;
+}
+
+void PCIeInfoHandler::deleteTopologyFiles()
+{
+    if (receivedFiles.empty())
+    {
+        try
+        {
+            for (auto& path : fs::directory_iterator(pciePath))
+            {
+                fs::remove_all(path);
+            }
+        }
+        catch (const fs::filesystem_error& err)
+        {
+            std::cerr << "Topology file deletion failed " << pciePath << " : "
+                      << err.what() << "\n";
+        }
+    }
 }
 
 } // namespace responder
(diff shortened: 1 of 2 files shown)


oem/ibm/libpldmresponder/file_io_type_pcie.hpp
```diff
@@ -285,6 +285,8 @@ class PCIeInfoHandler : public FileHandler
     static std::vector<std::string> cables;
     static std::vector<std::pair<linkId_t, linkId_t>> needPostProcessing;
     static std::unordered_map<linkId_t, linkType_t> linkTypeInfo;
+
```
Pavithrab7, line 288: Can you add a Doxygen on what this function does for future reference

oem/ibm/libpldmresponder/file_io_type_pcie.cpp
```diff
@@ -1105,5 +1106,24 @@ int PCIeInfoHandler::newFileAvailableWithMetaData(uint64_t /*length*/,
...
+            {
+                fs::remove_all(path);
+            }
+        }
+        catch (const fs::filesystem_error& err)
+        {
+            std::cerr << "Topology file deletion failed " << pciePath << " : "
+                      << err.what() << "\n";
+        }
+    }
```
Pavithrab7, line 1125: Do we want an else block on what happens when receivedFiles is not empty? Is this function called only when received files are empty?

ArchanaKakani: This function gets called for cable attributes file and topology file. We update received files after getting newFileAvailable for any of these files and we clear received files after receiving file ack for both the files.  So receivedFiles being empty is an indication to tell us that its a refresh(receiving new set of data) and you will get both the files one after the other. Else block is not required here as we need to delete cableinfo and topology file only if its a refresh.

oem/ibm/libpldmresponder/file_io_type_pcie.cpp
```diff
@@ -1105,5 +1106,24 @@ int PCIeInfoHandler::newFileAvailableWithMetaData(uint64_t /*length*/,
...
+{
+    if (receivedFiles.empty())
+    {
+        try
+        {
+            for (auto& path : fs::directory_iterator(pciePath))
+            {
+                fs::remove_all(path);
+            }
+        }
```
manojkiraneda, line 1119: why to use an iterator if you are using remove_all ?
can't we just do remove_all(pciePath) ?

manojkiraneda: and also remove_all() return 0 if path does not exist, do you wanna check that and return as well ? incase some one deletes that directory accidentally.

ArchanaKakani: remove_all(pciePath) will delete the directory with its content . To to avoid creation of directory multiple times I used iterator to just delete the directory content.

manojkiraneda: ack

manojkiraneda (approved):

sagisin (approved):

jaypadath (approved):

PR #319, opened 2022-10-18
PLDM: Changes to persist the HB PDRs in the repo before merging

Pavithrab7:
This commit persists the HB range PDRs in the repo before they are merged. This change was needed as PHYP would not send down BMC the entity Association PDRs of HB after a R/R as the PDRs will be in BMC range of record handles. Due to this the pldm ebmcModel in the PHYP was not updated properly.

This commit also prevents adding duplicate terminus locator PDRs to the repo.

Tested on both Rainier and Everest systems:
Multiple poweron and poweroff
Multiple Reset Reloads

Fixes: SW553100

Signed-off-by: Pavithra Barithaya <pavithra.b@ibm.com>

PR #323, opened 2022-10-28
Fix the wrong container id in SBE numeric effecter PDR

Pavithrab7:
This commit solves the issue caused by #319 where we persisted the HB entity association PDRs before we normalise them. This led to assigning the wrong container ids for the SBE numeric effecter PDRs created.
With this fix, we assign the normalised container ids to the numeric effecter PDRs.

This commit also removes the stale code to update the container
ids of proc LEDs. The JSON file is already updated with the
correct container id and this piece of code to update it is
not needed.

Fixes: PE00DXJZ


host-bmc/host_pdr_handler.cpp
```diff
@@ -6,6 +6,7 @@
 #include "libpldm/requester/pldm.h"
 #include "libpldm/state_set.h"
 #include "oem/ibm/libpldm/fru.h"
+#include "oem/ibm/libpldm/pdr_oem_ibm.h"
```
manojkiraneda, line 9: these two lines would be under #ifdef right?

sagisin (approved):

manojkiraneda (approved):

PR #324, opened 2022-11-04
oem-ibm : Code update FFDC enhancement

sagisin:
This commit adds additional tracing to print the
exception along with the error message in the we hit an exception while waiting for the Software Activation property.

### DIFF
--- a/oem/ibm/libpldmresponder/inband_code_update.cpp
+++ b/oem/ibm/libpldmresponder/inband_code_update.cpp
@@ -435,7 +435,10 @@ void CodeUpdate::setVersions()
                     }
                     catch (const sdbusplus::exception::exception& e)
                     {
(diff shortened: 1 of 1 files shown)


oem/ibm/libpldmresponder/inband_code_update.cpp
```diff
@@ -435,7 +435,9 @@ void CodeUpdate::setVersions()
                     }
                     catch (const sdbusplus::exception::exception& e)
                     {
-                        std::cerr << "Error in getting Activation status \n";
+                        std::cerr << "Error in getting Activation status,"
+                                  << "ERROR=" << e.what()
```
manojkiraneda, line 439: do you also wanna add other information? such as which interface, or object you are accessing that failed.

manojkiraneda (approved):

PR #335, opened 2023-01-05
DumpOffload: Clear SocketWriteStatus when Host poweroff

raviteja-b:
This commit clears SocketWriteStatus to "Free" when host power off
which avoids any wrong dump offload socket write status in  error cases

This PR fixes  race condition of status values across back to back MPIPLs 
when dump offload abruptly interrupted by power off 
https://w3.rchland.ibm.com/projects/bestquest/?defect=SW558457

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.hpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.hpp
@@ -114,6 +114,7 @@ class Handler : public oem_platform::Handler
                         setEventReceiverCnt = 0;
                         disableWatchDogTimer();
                         pldm::responder::utils::clearLicenseStatus();
+                        pldm::responder::utils::clearDumpSocketWriteStatus();
                     }
                     else if (propVal ==
                              "xyz.openbmc_project.State.Host.HostState.Running")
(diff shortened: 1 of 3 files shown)


sunharis (approved):

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
```diff
@@ -114,6 +114,7 @@ class Handler : public oem_platform::Handler
                         setEventReceiverCnt = 0;
                         disableWatchDogTimer();
                         pldm::responder::utils::clearLicenseStatus();
+                        pldm::responder::utils::clearDumpSocketWriteStatus();
```
manojkiraneda, line 117: Hi Ravi, Host goes to off state sometimes even in the middle of the power-on operation
- like when hostboot does a Reconfig loop.
- when we are doing an MPIPL as well.

Are we sure that in both cases - we are okay to clear the socket state?

raviteja-b: Manoj

yes we need to clear in both cases, whenever phyp stops running  then reset status flag

raviteja-b: does  this code hit in reconfigured loop path as well right?

manojkiraneda: no i guess, as its with hostboot

PR #340, opened 2023-01-18
Support New record to the backplane vpd for hostboot to consume

vkaverap:
This commit is to support a new VPD record called PSPD that contains few keywords called #D and VM. Since these keywords are typically larger in size (4K bytes) than what is managed by PLDM, we are adding a file into which these keywords will be extracted. This keyword file is transferred to the HB PLDM layer with a readFileByType request from HB.

Signed-off-by: Varsha Kaverappa <vkaverap@in.ibm.com>

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_by_type.cpp
+++ b/oem/ibm/libpldmresponder/file_io_by_type.cpp
@@ -2,6 +2,9 @@
 
 #include "file_io_by_type.hpp"
 
+#include <libpldm/base.h>
+#include <libpldm/file_io.h>
+
 #include "common/utils.hpp"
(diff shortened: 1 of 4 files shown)


oem/ibm/libpldmresponder/file_io_by_type.cpp
```diff
@@ -2,17 +2,17 @@
 
 #include "file_io_by_type.hpp"
 
-#include "libpldm/base.h"
-#include "libpldm/file_io.h"
-
```
manojkiraneda, line 7: why is this being moved down ?

manojkiraneda (approved):

jaypadath (approved):

PR #345, opened 2023-01-23
Modify Dump offload to use Thread

raviteja-b:
Currently while offloading a Dump pldmd writes on unix sockets 1M chunks When network is slow write to unix socket takes few mins which keeps pldmd busy with write() calls in a loop

This is implemented as discussed

PHYP sends the first 1MB chunk - BMC does the DMA operation, spin off's the thread and then return a PLDM_SUCCESS (after phyp gets the PLDM_SUCCESS they will unmap the TCE ) PHYP sends the second 1MB chuck - BMC would check the offload status and return PLDM_NOT_READY until the offload status is free Tested By:

1.Offload system dump
2.Offload system dump in slow networks speed

### DIFF
--- a/oem/ibm/libpldmresponder/file_io.cpp
+++ b/oem/ibm/libpldmresponder/file_io.cpp
@@ -19,6 +19,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <thread>
 
 namespace pldm
 {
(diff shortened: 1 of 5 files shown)


raviteja-b:
PLDM CI failed due to below error which is not related to this commit
[error] host-bmc/test/host_effecter_jsons/no_json/dummy.json: SyntaxError: Unexpected token (1:17)

PR #346, opened 2023-01-24
Support New record to the backplane vpd for hostboot to consume

vkaverap:
This commit is to support a new VPD record called PSPD that contains few keywords called #D and VM. Since these keywords are typically larger in size (4K bytes) than what is managed by PLDM, we are adding a file into which these keywords will be extracted. This keyword file is transferred to the HB PLDM layer with a readFileByType request from HB.

Signed-off-by: Varsha Kaverappa <vkaverap@in.ibm.com>


jaypadath (approved):

oem/ibm/libpldmresponder/file_io_type_vpd.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+namespace responder
+{
+int keywordHandler::read(uint32_t offset, uint32_t& length, Response& response,
+                         oem_platform::Handler* /*oemPlatformHandler*/)
+{
+    const char* keywrdObjPath =
+        "/xyz/openbmc_project/inventory/system/chassis/motherboard/vdd_vrm0";
+    const char* keywrdPropName = "PD_D";
+    const char* keywrdInterface = "com.ibm.ipzvpd.VPRI";
```
jaypadath, line 24: you need to use com.ibm.ipzvpd.PSPD instead of VPRI here

oem/ibm/libpldmresponder/file_io_type_vpd.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+namespace pldm
+{
+namespace responder
+{
+int keywordHandler::read(uint32_t offset, uint32_t& length, Response& response,
+                         oem_platform::Handler* /*oemPlatformHandler*/)
+{
+    const char* keywrdObjPath =
+        "/xyz/openbmc_project/inventory/system/chassis/motherboard/vdd_vrm0";
```
jaypadath, line 22: this path need to be "/xyz/openbmc_project/inventory/system/chassis/motherboard"

jaypadath (changes requested):

jaypadath (approved):

vkaverap:
gerrit commit - https://gerrit.openbmc.org/c/openbmc/libpldm/+/58013

vkaverap:
gerrit commit - https://gerrit.openbmc.org/c/openbmc/pldm/+/58403

manojkiraneda (approved):

PR #362, opened 2023-02-13
PLDM: Some more missed changes in 1050.

Pavithrab7:
This change was made in 1030 to skip the usage of container id for dimms in the state sensor events json file so that correct dimm will get deconfigued after applying a gauard operation.


manojkiraneda (approved):

Pavithrab7:
Without this change, it breaks the Hostboot DIMM guard functionality.

jaypadath (approved):

PR #372, opened 2023-03-01
Handle Dbus query exception gracefully

ArchanaKakani:
Handling the exception thrown by new_method_call

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -24,7 +24,6 @@ namespace pldm
 {
 namespace utils
 {
-
 constexpr auto mapperBusName = "xyz.openbmc_project.ObjectMapper";
 constexpr auto mapperPath = "/xyz/openbmc_project/object_mapper";
 constexpr auto mapperInterface = "xyz.openbmc_project.ObjectMapper";
@@ -246,12 +245,20 @@ GetSubTreeResponse
                             const std::vector<std::string>& ifaceList) const
 {
     auto& bus = pldm::utils::DBusHandler::getBus();
-    auto method = bus.new_method_call(mapperBusName, mapperPath,
-                                      mapperInterface, "GetSubTree");
-    method.append(searchPath, depth, ifaceList);
-    auto reply = bus.call(method);
-    GetSubTreeResponse response;
-    reply.read(response);
+    GetSubTreeResponse response{};
+    try
+    {
+        auto method = bus.new_method_call(mapperBusName, mapperPath,
+                                          mapperInterface, "GetSubTree");
+        method.append(searchPath, depth, ifaceList);
+        auto reply = bus.call(method);
+        reply.read(response);
+    }
+    catch (const std::exception& e)
+    {
+        std::cerr << "failed GetSubTree query for interface " << ifaceList[0]
+                  << " with ERROR=" << e.what() << "\n";
+    }
     return response;
 }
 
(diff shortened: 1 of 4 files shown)


oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -99,17 +99,17 @@ std::string DumpHandler::findDumpObjPath(uint32_t fileHandle)
...
     {
+        auto method =
+            bus.new_method_call(DUMP_MANAGER_BUSNAME, DUMP_MANAGER_PATH,
+                                OBJECT_MANAGER_INTERFACE, "GetManagedObjects");
         auto reply = bus.call(method);
         reply.read(objects);
     }
 
-    catch (const std::exception& e)
+    catch (const sdbusplus::exception_t& e)
```
sagisin, line 112: could you also make this change to other catch blocks in the file as well?

common/utils.cpp
```diff
@@ -246,12 +245,20 @@ GetSubTreeResponse
...
+                                          mapperInterface, "GetSubTree");
+        method.append(searchPath, depth, ifaceList);
+        auto reply = bus.call(method);
+        reply.read(response);
+    }
+    catch (const std::exception& e)
+    {
+        std::cerr << "failed GetSubTree query for interface " << ifaceList[0]
+                  << " with ERROR=" << e.what() << "\n";
+    }
```
manojkiraneda, line 261: no, this is not the right place to catch the exception, any function who is call this getSubtree should have the function call in try - and - catch block. so that at the the function call - we can decide what to do with the generated exception.

manojkiraneda (changes requested):

common/utils.cpp
ArchanaKakani: GetSubTreeResponse response{} is an empty vector. For any exception we will log an error and return empty vector. The function which calls getSubtree will to traverse the vector and takes action on the data present in the vector. So catching the exception here in the util function should be fine.

manojkiraneda: what would happen to this case where , the client code that will call this function need to know if the dbus call failed vs the dbus call returned 0 dbus objects ? and take the necessary action based on it ? 

Right now with this code, the client will always receive {} when the call failed or even when the number of objects found by mapper are 0.

ArchanaKakani: I think in both the cases action remains the same. Do you have any specific case were you think this should be handled seperately

oem/ibm/libpldmresponder/file_io_type_dump.cpp
ArchanaKakani: done

common/utils.cpp
manojkiraneda: i don't have any use cases in mind, but I think the community would not agree with this way of designing, we should always give the client an option to choose how they want to handle the generated exceptions, rather than deciding them in the utility function.

manojkiraneda (approved):

Pavithrab7 (approved):

jaypadath (approved):

common/utils.cpp
ArchanaKakani: For now, handling the exception in util function only as we are not taking any specific action on failure in caller function.

PR #389, opened 2023-03-27
Listen for ifacesAdded for BIOS attribute props

spinler:
PLDM was just watching for PropertiesChanged signals on the D-Bus properties that back BIOS attributes.  Just doing that can miss properties on interfaces that get added after PLDM starts and then never change after that, because they would just send a single InterfacesAdded signal.

Close that hole by also adding an InterfacesAdded match along with the PropertiesChanged match.

Change-Id: Iab6679b0c99d049bd5f11e6508f784b2e8eb9349


spinler:
Upstream is https://gerrit.openbmc.org/c/openbmc/pldm/+/61916
Defect 460992

manojkiraneda (approved):

PR #401, opened 2023-05-02
Adding bios attributes for RPD feature

Patel-Kamalkumar:
- Adding bios attributes for runtime processor diagnostics
feature which allow HOST to consume bios attribute at
runtime and after that it's depend on HOST decision to
apply immediately  or next IPL cycle.

### DIFF
--- a/oem/ibm/configurations/bios/integer_attrs.json
+++ b/oem/ibm/configurations/bios/integer_attrs.json
@@ -196,6 +196,15 @@
             "displayName": "RPD Scheduled Run Time of Day in Seconds"
         },
         {
+            "attribute_name": "pvm_rpd_scheduled_duration",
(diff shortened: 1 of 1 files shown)


geissonator:
Latest CI doesn't like that there's no newline at end of file:
```
[1m+++ b/oem/ibm/configurations/bios/integer_attrs.json[m
[36m@@ -269,4 +269,4 @@[m
             }[m
         }[m
     ][m
[31m-}[m
\ No newline at end of file[m
[32m+[m[32m}[m
Format: [31mFAILED[0m
```

Patel-Kamalkumar:
I did not add any new line, i think while creating PR it got added automatically. 
same issue observed with one more team member recently, let me check with her.

PR #411, opened 2023-05-19
systemd: no installation in templated targets

geissonator:
Upstream yocto introduced a change via e510222 (systemd-systemctl: fix instance template WantedBy symlink construction).

This fixes a bug that we in OpenBMC had been taking advantage of in that we were able to document our templated target dependencies without it actually doing anything. The real installation of services within targets occurs in our bitbake recipes due to the complexity of chassis and host instances on a per machine basis.

Leave the dependency information in the service files but comment them out. It's useful to be able to look at a service and understand which targets it's going to be installed into by the bitbake recipes.

In some cases, we had hard coded the target instance, which does install the service correctly, but only in that one target. All services should be installed via the bitbake recipe to ensure the service is properly installed in all instances of the target. Once the bump for this commit goes into openbmc/openbmc, I will ensure the recipe is updated to install all services correctly.

### DIFF
--- a/oem/ibm/service_files/pldm-clear-dma.service
+++ b/oem/ibm/service_files/pldm-clear-dma.service
@@ -8,8 +8,8 @@ Conflicts=obmc-host-startmin@0.target
 [Service]
 RemainAfterExit=yes
 Type=oneshot
-ExecStart=/usr/bin/killall -SIGSTOP pldmd 
-ExecStart=/usr/bin/killall -SIGCONT pldmd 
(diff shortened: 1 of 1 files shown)


geissonator:
This is currently a downstream-only service file so need to fix in downstream (no upstream gerrit). This same change was done in a variety of upstream repos via https://gerrit.openbmc.org/q/topic:systemd-templates

manojkiraneda (approved):

PR #442, opened 2023-07-20
Return proper Success/Error code to host

sagisin:
This commit adds change to send the right error code to host during the marker lid validation. When the marker lid validation fails we were trying to send a PLDM_ERROR But it was sent as a response to the file write request and PHYP assumed that the error code returned was for the write operation and did not go ahead to check the sensor event sent to indicate the entitlment fail during marker lid validation.

Defect: PE00LNC2

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_lid.hpp
+++ b/oem/ibm/libpldmresponder/file_io_type_lid.hpp
@@ -118,10 +118,8 @@ class LidHandler : public FileHandler
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


anoo1 (approved):

anoo1:
Sagar tested and a detailed message now appears on the OS:

```
The selected firmware image can not be applied.

The Build Date of the firmware image selected is
    June 01, 2024.
The System's Update Access Key Expiration Date is
    December 31, 2023.

Please contact your service support structure.
```

jaypadath (approved):

PR #444, opened 2023-08-07
PLDM: 5 sec timeout for dbus calls without reply

vkaverap:
This commit adds 5 second dbus timeout value to dbus calls with no reply. Previously a commit was merged that included 5 seconds dbus timeout value for all dbus call with reply.

This is added to make sure PLDM does not wait for more than 5 seconds when a dbus call is made with or without reply if no response is received from other daemon or host.

The 5 second  timeout value is overridden to 10 seconds in meta-ibm layer

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -289,7 +289,9 @@ void reportError(const char* errorMsg, const Severity& sev)
         }
         std::map<std::string, std::string> addlData{};
         method.append(errorMsg, severity, addlData);
-        bus.call_noreply(method);
+        bus.call_noreply(
+            method,
+            std::chrono::duration_cast<microsec>(sec(DBUS_TIMEOUT)).count());
     }
     catch (const std::exception& e)
     {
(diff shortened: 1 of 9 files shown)


jaypadath (commented):
Hi Varsha, you mention of 5 second time out is confusing. Functionally with this change we should have 10sec timeout for all the noreply dbus calls. Can you please update the commit message accordingly ?

common/utils.cpp
```diff
@@ -343,7 +347,9 @@ void DBusHandler::setDbusProperty(const DBusMapping& dBusMap,
             }
             method.append(dBusMap.interface.c_str(),
                           dBusMap.propertyName.c_str(), variant);
-            bus.call_noreply(method);
+            bus.call_noreply(
+                method, std::chrono::duration_cast<microsec>(sec(DBUS_TIMEOUT))
```
ArchanaKakani, line 351: Rather than casting DBUS_TIMEOUT every time, can we set it to one common variable and use it?

riyadixitagra (approved):

common/utils.cpp
vkaverap: wil check if we can add it in utils.hpp

jaypadath (approved):

common/utils.cpp
vkaverap: this will be addressed in subsequent commits since the suggested change also has to be made in files that are not part of this commit

PR #446, opened 2023-08-17
Throw Marker Lid validate error

sagisin:
Commit the marker lid validate errors for the
min-mif level and access key expiration so that
it generates an SRC when we hit the error during
code update.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_lid.hpp
+++ b/oem/ibm/libpldmresponder/file_io_type_lid.hpp
@@ -1,7 +1,10 @@
 #pragma once
 
 #include "file_io_by_type.hpp"
+#include "xyz/openbmc_project/Common/error.hpp"
 
+#include <phosphor-logging/elog-errors.hpp>
+#include <phosphor-logging/elog.hpp>
 #include <phosphor-logging/lg2.hpp>
 #include <xyz/openbmc_project/Software/Version/error.hpp>
 
@@ -21,9 +24,13 @@ using MarkerLIDremainingSize = uint64_t;
 
 constexpr auto markerLidName = "80a00001.lid";
 constexpr auto accessKeyExpired =
-    "sdbusplus::xyz::openbmc_project::Software::Version::Error::ExpiredAccessKey";
+    "xyz.openbmc_project.Software.Version.Error.ExpiredAccessKey";
+using AccessKeyExpired =
+    sdbusplus::xyz::openbmc_project::Software::Version::Error::ExpiredAccessKey;
 constexpr auto incompatibleErr =
-    "sdbusplus::xyz::openbmc_project::Software::Version::Error::Incompatible";
+    "xyz.openbmc_project.Software.Version.Error.Incompatible";
+using IncompatibleErr =
+    sdbusplus::xyz::openbmc_project::Software::Version::Error::Incompatible;
 
 /** @class LidHandler
  *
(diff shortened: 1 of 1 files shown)


oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -22,8 +25,10 @@ using MarkerLIDremainingSize = uint64_t;
 constexpr auto markerLidName = "80a00001.lid";
 constexpr auto accessKeyExpired =
     "sdbusplus::xyz::openbmc_project::Software::Version::Error::ExpiredAccessKey";
```
anoo1, line 27: This string should be changed to: "xyz.openbmc_project.Software.Version.Error.ExpiredAccessKey";
That's the string name of the exception from the generated PDI Version/error.hpp file:
static constexpr auto errName = "xyz.openbmc_project.Software.Version.Error.ExpiredAccessKey";

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -22,8 +25,10 @@ using MarkerLIDremainingSize = uint64_t;
...
+using AccessKeyExpired = sdbusplus::xyz::openbmc_project::Software::Version::Error::ExpiredAccessKey;
 constexpr auto incompatibleErr =
     "sdbusplus::xyz::openbmc_project::Software::Version::Error::Incompatible";
```
anoo1, line 30: This string should be changed to: "xyz.openbmc_project.Software.Version.Error.Incompatible";
That's the string name of the exception from the generated PDI Version/error.hpp file:
static constexpr auto errName = "xyz.openbmc_project.Software.Version.Error.Incompatible";

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -150,10 +155,12 @@ class LidHandler : public FileHandler
             {
                 if (strcmp(e.name(), accessKeyExpired) != 0)
```
anoo1, line 156: The check should be for a return value of 0:
if (strcmp(e.name(), accessKeyExpired) == 0)

strcmp() returns 0 if the contents of both strings are equal.

0 | the contents of both strings are equal
-- | --

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -150,10 +155,12 @@ class LidHandler : public FileHandler
...
                 {
+                    phosphor::logging::commit<AccessKeyExpired>();
                     validateStatus = ENTITLEMENT_FAIL;
                 }
                 else if (strcmp(e.name(), incompatibleErr) != 0)
```
anoo1, line 161: The check should be for a return value of 0:
if (strcmp(e.name(), accessKeyExpired) == 0)

strcmp() returns 0 if the contents of both strings are equal.

anoo1 (changes requested):
Found a bug that was already present while trying to test these changes to create an error with the min mif value. I think it can be corrected with this commit since this commit adds the ability to create an error log for the 2 different scenarios, and without the proposed changes, it'll always create the ExpiredAccessKey error only.

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: Done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: Done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: Done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: Done

jaypadath (approved):

PR #447, opened 2023-08-21
pdr: Missing nvme slot entries in the json

Pavithrab7:
According to the rainier workbook, the rainier1s-4u model has 16 nvme slots 8 in each disk_backplane0 and disk_backplane1.
The ibm,rainier1s-4u json had missing nvme slots entries and were not modelled by PLDM. This commit adds the same.

Fixes: 561341

Change-Id: I0559f1761a6ec223fa62aa14c283885d3aa37e49

### DIFF
--- a/oem/ibm/configurations/pdr/ibm,rainier-1s4u/11.json
+++ b/oem/ibm/configurations/pdr/ibm,rainier-1s4u/11.json
@@ -615,6 +615,44 @@
                     ]
                 },
                 {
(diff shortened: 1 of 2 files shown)


jaypadath (changes requested):
Since we are updating the slots, we need to update the power states also (State Set Id 257 : xyz.openbmc_project.State.Decorator.PowerState) for newly added slots

Pavithrab7:
> Since we are updating the slots, we need to update the power states also (State Set Id 257 : xyz.openbmc_project.State.Decorator.PowerState) for newly added slots

>>Done

jaypadath (approved):

PR #461, opened 2023-09-13
oem-ibm: Call PEL API when host drops event log

spinler:
When IBM's hypervisor receives an event log (PEL) using the file API, it may end up discarding it, returning the reason the file was discarded in the status argument of the file ack command.  It can do this for two reasons, either it doesn't have any more room to store event logs, or the event log was malformed.

In these cases, call the HostReject D-Bus method on the PEL daemon so it can handle it. The code must convert the value of the status returned into the proper D-Bus enum value.

Without this call, the operating system was missing event logs when there were floods of them.

Tested:
API was called when system was flooded with event logs, allowing the PEL daemon to resend them as space opens up.

Change-Id: I9d3acf4e5c5f885934ef96e89ddb93dd8f9c10a3

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_pel.cpp
+++ b/oem/ibm/libpldmresponder/file_io_type_pel.cpp
@@ -10,6 +10,7 @@
 #include <systemd/sd-bus.h>
 #include <unistd.h>
 
+#include <org/open_power/Logging/PEL/server.hpp>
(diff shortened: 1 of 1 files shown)


spinler:
Upstream: https://gerrit.openbmc.org/c/openbmc/pldm/+/66566

jaypadath (approved):

PR #462, opened 2023-09-20
Removing the dead code

ArchanaKakani:
executeCmd() function is not getting used anywhere in the code.

Usage of executeCmd function was removed as part of inband code update changes.
 Now its no longer used so removing it.


jaypadath (commented):
Please reduce and rephrase the title (executeCmd() function is not getting used anywhere in the code) length and add "oem_ibm: " at beginning of the commit title.

Reference openbmc doc https://github.com/openbmc/docs/blob/master/CONTRIBUTING.md#submitting-changes on title length suggestion

jaypadath (approved):

sagisin (approved):

PR #463, opened 2023-09-21
PLDM :Shorten typecast expression for DBUS_TIMEOUT

vkaverap:
This commits uses constexpr to shorten the length of DBUS_TIMEOUT sent as an input parameter after tyecasting it to microseconds for bus.call calls.


ArchanaKakani (commented):
In the commit msg , it is good to mention that configured Dbus timeout value is in seconds and we are creating a global variable to hold the converted microseconds which will be used in all the bus.call calls.

ArchanaKakani (approved):

PR #465, opened 2023-09-26
PLDM :Shorten typecast expression for DBUS_TIMEOUT

vkaverap:
This commit adds a global variable to hold the convertion of DBUS_TIMEOUT value to microseconds from seconds. Constexpr is used to shorten the length of DBUS_TIMEOUT sent as an input parameter after typecasting it to microseconds for all bus.call calls.


ArchanaKakani (commented):
The timeout changes are made for bus.call and bus.callnoreply, update the commit msg accordingly. And reframe the second sentence to by providing only needed information, not sure if typecasting details are needed here.

ArchanaKakani (approved):

jaypadath (approved):

PR #470, opened 2023-10-06
Responder support for Get numeric effecter command

ArchanaKakani:
Added responder support for GetNumericEffecter Command

Test:
Using pldmtool raw command
    pldmtool raw -d 0x80 0x02 0x32 0x23 00
    pldmtool: Tx: 08 01 80 02 32 23 00
    pldmtool: Rx: 08 01 00 02 32 00 00 01 03 03

### DIFF
--- a/libpldmresponder/platform_numeric_effecter.hpp
+++ b/libpldmresponder/platform_numeric_effecter.hpp
@@ -333,6 +333,222 @@ int setNumericEffecterValueHandler(const DBusInterface& dBusIntf,
     return PLDM_SUCCESS;
 }
 
+/** @brief Function to convert the D-Bus value by effecterdataSize and create
+ * the response
+ *  @for getNumericEffecterValue request.
+ *  @param[in] PropertyValue - D-Bus Value
+ *  @param[in] effecterDataSize - effecter value size.
+ *  @param[in,out] responsePtr - Response of getNumericEffecterValue.
+ *  @param[in] responsePayloadLength - reponse length.
+ *  @param[in] instanceId - instance id for response
+ *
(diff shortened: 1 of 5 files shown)


libpldmresponder/platform_numeric_effecter.hpp
```diff
@@ -333,6 +336,162 @@ int setNumericEffecterValueHandler(const DBusInterface& dBusIntf,
...
 }
 
+int getEffecterValue(std::string propertyType,
+                     pldm::utils::PropertyValue propertyValue,
+                     uint8_t* effecterValue, size_t effecterValueLength)
+{
+    if (propertyType == "uint32_t")
+    {
+        uint32_t propVal = std::get<uint32_t>(propertyValue);
+#ifdef __STDC_LIB_EXT1__
```
Pavithrab7, line 346: why a C style copying, can't we use casting and then return the value?

jaypadath: Please try of std::copy_n works here instead of memcpy

Pavithrab7: please respond to the above comments

ArchanaKakani: Removed memcpy 

> why a C style copying, can't we use casting and then return the value?

Removed the memcpy and directly passing the effecter value by typecasting it to reinterpret_cast.

ArchanaKakani: > please respond to the above comments

I have reworked on the code flow for GetNumericEffecter. Now we are not using memcpy. Requesting review again.

jaypadath (approved):

PR #472, opened 2023-10-12
PLDM : Change the dbus timeout to 5 seconds

vkaverap:
Dbus timeout value for every dbus call in PLDM is set to 5 seconds instead of default 25 seconds so that all dbus calls that expect a reply exits after 5 seconds if there is no response from the client. While BMC is still waiting for a response on any dbus calls it is possible that host can timeout and return an error. Another reason is that memory address from where BMC would read data sent by host may get deallocated after 20 seconds and it is possible for this to happen even before BMC has read the data. This has lead to DMA issue in other releases.


ArchanaKakani (commented):
typo exits

ArchanaKakani (commented):
How is the DMA issue related to Dbus timeout?

vkaverap:
> How is the DMA issue related to Dbus timeout?

please read the description in the main storu for dbus timeout

jaypadath (approved):

PR #473, opened 2023-10-12
Adding response mechanism to remote terminus

Patel-Kamalkumar:
- File transfer between BMC and Host is synchronous mechanism.
- File transfer interrupted because of many reasons, so if
  file transfer interrupted between BMC and Host then PLDM is
  ending up in the hung state and not able to respond to the
  upcoming new requests. User has to do work around to restart
  the PLDM to get work again.
- I have added asynchronous file transfer support using
  eventloop mechanism and non-blocking socket communication,
  so file transfer will be taken care by event loop and PLDM
  is free to receive another request from the Host.
- If file transfer will be stuck or interrupted during
  asynchronous transfer then time-out occurs and file transfer
  will be aborted and same status communicated to the Host.

### DIFF
--- a/common/utils.hpp
+++ b/common/utils.hpp
@@ -61,11 +61,13 @@ struct CustomFD
     CustomFD(CustomFD&&) = delete;
     CustomFD& operator=(CustomFD&&) = delete;
 
-    CustomFD(int fd) : fd(fd) {}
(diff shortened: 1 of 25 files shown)


manojkiraneda (approved):
Looks good to me.

jaypadath (approved):

PR #488, opened 2023-12-04
oem: ibm: fileTable: Add lid 81e00670

anoo1:
Add read-only lid 81e00670 which is a partition table that hostboot uses to get information about the other partition files.

Tested: Verified this lid was read by hostboot on a p10bmc system and that the system powered on.

Change-Id: If09ec0ae0938925f8a147dc28ea2d8b24f94c13d

### DIFF
--- a/oem/ibm/configurations/fileTable.json
+++ b/oem/ibm/configurations/fileTable.json
@@ -84,6 +84,10 @@
         "file_traits": 1
     },
     {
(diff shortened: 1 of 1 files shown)


anoo1:
Upstream: https://gerrit.openbmc.org/c/openbmc/pldm/+/68069

sagisin (approved):

PR #489, opened 2023-12-05
oem_ibm:Adding vmi_if_count bios attribute

isha309:
-This attribute is needed by
 phosphor-networkd to identify
 total number of ethernet interfaces

Change-Id: I804b120086ab30f4579de830ba64ea358f059987


manojkiraneda:
This is no longer required as we have system specific bios attributes support.

PR #504, opened 2024-01-03
oem-ibm: Disabling Active Memory Mirroring for Bonnell

ArchanaKakani:
Active memory mirroring is not supported in Bonnell. To remove that support
we are removing the bios attribute completely so that user will not have any access
to use the AMM setting.
Removed the bios attribute hb_memory_mirror_mode and hb_memory_mirror_mode_current

Fixes: 588132

Tested:
Poweron/poweroff and reset reload case, verified
   /redfish/v1/Registries/BiosAttributeRegistry/BiosAttributeRegistry

### DIFF
--- a/oem/ibm/configurations/bios/ibm,bonnell/enum_attrs.json
+++ b/oem/ibm/configurations/bios/ibm,bonnell/enum_attrs.json
@@ -273,21 +273,6 @@
             "displayName": "Secure Version Lockin Enabled"
         },
         {
-            "attribute_name": "hb_memory_mirror_mode",
-            "possible_values": ["Disabled", "Enabled"],
-            "default_values": ["Disabled"],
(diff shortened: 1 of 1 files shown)


jaypadath (commented):
Please change the commit title "pldm" to "oem-ibm". Also add some more info in to the commit message description on why we are removing it.

vkaverap:
> Please change the commit title "pldm" to "oem-ibm". Also add some more info in to the commit message description on why we are removing it.

jaypadath (commented):
Remove "oem-ibm: Disabling Active Memory Mirroring for Bonnell" from commit body as that will be redundant info.

jaypadath (approved):

vkaverap (approved):

PR #509, opened 2024-01-30
oem-ibm: Multipath Splitter support

ArchanaKakani:
Added below changes to support multipath splitter

1. Add LinkReset to Logical slots
2. Fetch Logical slot entities from saved object paths.
3. Updated code to fetch the location code of logical slots too.

Tested:
1. Inventory LEDs page on GUI.
2. Identifying LEDs on Inventory Page works well.
3. Topology Page (Populates links properly)
4. Link Reset on few links
5. Verified Dbus details like Location code of Logical slots and Dbus Modelling of Topology data.

### DIFF
--- a/host-bmc/host_pdr_handler.cpp
+++ b/host-bmc/host_pdr_handler.cpp
@@ -1827,7 +1827,7 @@ void HostPDRHandler::createDbusObjects()
             case PLDM_ENTITY_SLOT:
                 CustomDBus::getCustomDBus().implementPCIeSlotInterface(
                     entity.first);
-                CustomDBus::getCustomDBus().setlinkreset(
+                CustomDBus::getCustomDBus().setLinkReset(
                     entity.first, false, hostEffecterParser, mctp_eid);
                 break;
             case PLDM_ENTITY_CONNECTOR:
@@ -1856,6 +1856,8 @@ void HostPDRHandler::createDbusObjects()
                 CustomDBus::getCustomDBus().setSlotType(
                     entity.first,
                     "xyz.openbmc_project.Inventory.Item.PCIeSlot.SlotTypes.OEM");
+                CustomDBus::getCustomDBus().setLinkReset(
+                    entity.first, false, hostEffecterParser, mctp_eid);
                 break;
             default:
                 break;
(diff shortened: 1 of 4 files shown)


oem/ibm/libpldmresponder/file_io_type_pcie.cpp
```diff
@@ -206,10 +206,11 @@ void PCIeInfoHandler::getMexObjects()
     // get the in memory cached copy of mex objects
     auto savedObjs = pldm::serialize::Serialize::getSerialize().getSavedObjs();
 
-    // Find the PCIE slots & PCIeAdapters & Connecters & their location codes
+    // Find the PCIE slots & PCIE logical slots & PCIeAdapters & Connecters &
+    // their location codes
     std::set<uint16_t> neededEntityTypes = {PLDM_ENTITY_SLOT, PLDM_ENTITY_CARD,
                                             PLDM_ENTITY_CONNECTOR,
-                                            PLDM_ENTITY_SYSTEM_CHASSIS};
+                                            PLDM_ENTITY_SYSTEM_CHASSIS, 32954};
```
Pavithrab7, line 213: magic number?

host-bmc/host_pdr_handler.cpp
```diff
@@ -1856,6 +1856,8 @@ void HostPDRHandler::createDbusObjects()
                 CustomDBus::getCustomDBus().setSlotType(
                     entity.first,
                     "xyz.openbmc_project.Inventory.Item.PCIeSlot.SlotTypes.OEM");
+                CustomDBus::getCustomDBus().setlinkreset(
```
Pavithrab7, line 1859: setLinkReset - camel case usage?

oem/ibm/libpldmresponder/file_io_type_pcie.cpp
ArchanaKakani: This is the logical slot entity Id , we have used the same at multiple places so adding the same to avoid confusion

host-bmc/host_pdr_handler.cpp
ArchanaKakani: This needs changes at multiple other places, can we take this as separate commit?

oem/ibm/libpldmresponder/file_io_type_pcie.cpp
Patel-Kamalkumar: why can't we add ENUM string instead of magic number? 
and readability is also problem incase of magic number, if we don't have any mechanism yet then need to think to avoid magic numbers in code.

manojkiraneda: we can change it to 0x8000 | PLDM_ENTITY_SLOT

manojkiraneda (changes requested):

oem/ibm/libpldmresponder/file_io_type_pcie.cpp
ArchanaKakani: Done

jaypadath (approved):

Pavithrab7 (approved):

host-bmc/host_pdr_handler.cpp
ArchanaKakani: Done

PR #518, opened 2024-02-20
pldm: SBE Dump Support for Odyssey fix

riyadixitagra:
This commit addresses some minor
issues discovered as part of integration
testing. Issues addressed:
- Invalid entity type
- Notification being sent to HB for all OCMB instead of the only one OCMB which is completing the dump. This is fixed for PROCs too.

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.cpp
@@ -105,8 +105,8 @@ int pldm::responder::oem_ibm_platform::Handler::
 
         else
         {
-            error("Invalid entity type received. Entity type = {ENTITY_TYP}",
-                  "ENTITY_TYP", entityType);
+            error("Invalid entity type received: {ENTITY_TYPE}", "ENTITY_TYPE",
+                  entityType);
         }
         rc = setNumericEffecter(entityInstance, value, entityType);
     }
@@ -999,7 +999,7 @@ int encodeEventMsg(uint8_t eventType, const std::vector<uint8_t>& eventDataVec,
     return rc;
 }
 void pldm::responder::oem_ibm_platform::Handler::setHostEffecterState(
-    bool status, uint16_t entityTypeReceived)
+    bool status, uint16_t entityTypeReceived, uint16_t entityInstance)
 {
     pldm::pdr::EntityType entityType;
     if (entityTypeReceived == PLDM_ENTITY_MEMORY_MODULE)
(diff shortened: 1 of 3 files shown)


oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -105,8 +105,9 @@ int pldm::responder::oem_ibm_platform::Handler::
 
         else
         {
-            error("Invalid entity type received. Entity type = {ENTITY_TYP}",
-                  "ENTITY_TYP", entityType);
+            error(
+                "Invalid entity type received by function oemSetNumericEffecterValueHandler. Entity type = {ENTITY_TYP}",
+                "ENTITY_TYP", entityType);
```
Patel-Kamalkumar, line 110: can we start logging error traces as per andrews comment from here onwards?
please go through all comment and apply here.
[](https://gerrit.openbmc.org/c/openbmc/pldm/+/67994)

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1026,88 +1028,92 @@ void pldm::responder::oem_ibm_platform::Handler::setHostEffecterState(
...
+                    return;
+                }
+                uint8_t completionCode{};
+                auto rc = decode_set_state_effecter_states_resp(
+                    response, respMsgLen, &completionCode);
+                if (rc)
+                {
+                    error(
+                        "Failed to decode setStateEffecterStates response,rc= {RC}",
+                        "RC", static_cast<unsigned>(rc));
```
Patel-Kamalkumar, line 1083: please remove static_cast, it's not required since LG2 supports uint8.

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -1012,8 +1013,9 @@ void pldm::responder::oem_ibm_platform::Handler::setHostEffecterState(
     }
     else
     {
-        error("Invalid entity type received. Entity Type = {ENTITY_TYP}",
-              "ENTITY_TYP", entityTypeReceived);
+        error(
+            "Invalid entity type received by function setHostEffecterState. Entity Type = {ENTITY_TYP}",
```
Patel-Kamalkumar, line 1017: no need to write function name inside trace "setHostEffecterState".

riyadixitagra: there are three places where we have same error message being printed due to which we took some extra time to debug so based on discussion with @Pavithrab7 this change was made.

riyadixitagra: I have removed the entity type written in message but the other thing of using function name in lg2, have you seen any instance? I am aware of this thing working for std::cout but lg2 I am doubtful.

riyadixitagra: will be omitting the file names.

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
riyadixitagra: Done

Patel-Kamalkumar (approved):
Looks good to me.

Pavithrab7 (approved):

jaypadath (approved):

PR #521, opened 2024-03-10
pldm: System specific bios attributes support with  lazy BIOS Table

ArchanaKakani:
Entity manager service hosts the compatible interface and populates the Names property based on probing the IM Keyword. And it is seen that the Names property is updated pretty late on the way to BMC ready state. And by that time pldmd is already creating the bios table assuming that the system type is not detected.

To fix this behavior, rather than creating the bios table early, we need to wait until we get the system type from the entity manager daemon.

System Type is fetched from Entity Manager
Decorator.Compatible interface.

PLDM registers the service name only after building the BIOS tables.

Added meson option "system-specific-json" to add the support for system specific bios attributes.
Below is the pre-requisite to support system specific bios attributes:
1. Entity Manager service is active
2. Entity Manager should have Decorator.Compatible interface and system type in Names property under this interface
3. BIOS Attribute json files are added into the folder(Named on the System Type property value) and should be installed under /usr/share/pldm/bios/

With disabled "system-specific-json" option default bios attributes are populated.

As Bios table is build lazily during genesis boot bootside file is getting written with empty current boot side as well as next boot side. Updated the bootside file with default boot side as "temp" during genesis boot.

Tested:
      1. Power ON/OFF
      2. Reset Reload
      3. Jenkin's mutiple boot test
      4. Factory Reset


Pavithrab7:
are all changes same as upstream? did we make any other change only with respect to downstream?

ArchanaKakani:
> are all changes same as upstream? did we make any other change only with respect to downstream?
Changes are almost same as upstream

jaypadath (approved):
Changes looks fine to me. Also tested section covers most of the scenarios.

Pavithrab7 (commented):
Do you want to split the commit into 2? 
1. Add support for moving the compatible interface out of oem-ibm and making use of generic interface?
2. Lazy Bios Support using the same interface?

I am okay anyway, as this is a huge PR :)

PR #525, opened 2024-04-25
oem-ibm: Fixing DMA length issue

Patel-Kamalkumar:
When HOST read/write data to BMC using DMA then BMC checks the length of the data to be read/write and devide into chunks if data length is more than DMA max size.
This change will fix the issue in PLDM when it send data transfer status to the HOST and sending left over length as response which is wrong and now PLDM sends correct response with length.
_Defect_: https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=612662

Tested:
IPL flow tested and PHYP reaches till standby which include all type of file transfer from BMC to PHYP.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_by_type.hpp
+++ b/oem/ibm/libpldmresponder/file_io_by_type.hpp
@@ -42,7 +42,7 @@ class FileHandler
      *                                  tasks
      *  @return PLDM status code
      */
-    virtual int readIntoMemory(uint32_t offset, uint32_t& length,
+    virtual int readIntoMemory(uint32_t offset, uint32_t length,
                                uint64_t address,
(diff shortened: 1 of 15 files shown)


jaypadath (commented):
Add oem-ibm: with Git Title (Fixing DMA length issue). Add the tested scenarios.

sagisin:
Can you add the defect and slack thread reference to the commit message for future reference?

sagisin (approved):

jaypadath (approved):

manojkiraneda:
changes look okay to me. But can you also push this upstream please before we merge this downstream ? i don't think this is a downstream only solution right ?

manojkiraneda (approved):

PR #530, opened 2024-05-07
1020: Fixing Read lid size corruption issue (#525)

rfrandse:
#### Fixing Read lid size corruption issue (#525)
```
When HOST read/write data to BMC using DMA then BMC checks the length
of the data to be read/write and divide into chunks if data length is
more than DMA max size.
This change will fix the issue in PLDM when it send data transfer
status to the HOST and sending left over length as response which is
wrong and now PLDM sends correct response with length.

Signed-off-by: Kamalkumar Patel <kamalkumar.patel@ibm.com>
```


rfrandse:
jenkins run tests please

PR #537, opened 2024-05-20
host-bmc: Update set host effecter PEL severity

riyadixitagra:
This commit updates the severity of PEL
'xyz.openbmc_project.PLDM.Error.SetHostEffecterFailed' from ERROR to INFORMATIONAL.

Test: Triggered set numericEffecter at runtime and verified the severity of PEL logged as INFORMATIONAL.


jaypadath (approved):

geissonator (commented):
General comment to please include the "why" in the commit message. So when we come look at this in a year we know why we went from Error to Info on this error.

PR #539, opened 2024-05-28
oem-ibm: Adding BIOS attribute for dynamic deallocation of Memory

riyadixitagra:
This commit adds a BIOS attribute for allowing user to enable/disable the dynamic deallocation of memory.

Tested: Verified the BIOS attribute with get bios attribute command from pldmtool.


ArchanaKakani (commented):
Do we need to add one more bios attribute `hb_predictive_mem_guard_current` as changes to `hb_predictive_mem_guard` bios attribute takes effect after next reboot?

Pavithrab7 (changes requested):
We need a _current attribute too, please check with the HB team and update.

Pavithrab7 (approved):

Pavithrab7:
can you add a comment about what is dynamic memory allocation in the commit message?

PR #557, opened 2024-06-14
hotplug: Add API to check FRU presence

vkaverap:
API that checks if FRU with a given object path is present in the inventory interface is introduced. FRU records are built only for FRUs that are present in the inventory interface list. The presence of a FRU is determined by the "Present" Dbus property.

Tested By: IPL and reset on BMC.

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -659,5 +659,25 @@ void setFruPresence(const std::string& fruObjPath, bool present)
     }
 }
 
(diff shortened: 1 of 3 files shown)


Pavithrab7:
Varsha, checkFRUPresence API is already merged upstream and is also present in 1110. The changes are already present.

I guess this PR is redundant.

vkaverap:
yes realised that

PR #571, opened 2024-07-10
pldm: Adding Fix for NVRAM checksum corruption

Patel-Kamalkumar:
When user do chasisoff then power is abruptly down for chassis and NVRAM file transfer is abruptly stops which leads to mismatch NVRAM file checksum and PHYP updated checksum due to data corruption. This commit adds the synchronous file transfer mechanism only for NVRAM and Checksum file to avoid race condition while chassis off.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io.hpp
+++ b/oem/ibm/libpldmresponder/file_io.hpp
@@ -122,13 +122,23 @@ class DMA
     /** @brief Method to fetch the shared memory file descriptor for data
      * transfer
      *
+     * @param[in] isNonBlckSocket - Varible to decide socket should be blocking
+     * / non blocking mode while opening
+     *
      *  @return returns shared memory file descriptor
      */
-    int getNewXdmaFd()
+    int getNewXdmaFd(bool isNonBlckSocket = true)
     {
         try
         {
-            xdmaFd = open(xdmaDev, O_RDWR | O_NONBLOCK);
+            if (isNonBlckSocket)
+            {
+                xdmaFd = open(xdmaDev, O_RDWR | O_NONBLOCK);
+            }
+            else
+            {
+                xdmaFd = open(xdmaDev, O_RDWR);
+            }
         }
         catch (...)
         {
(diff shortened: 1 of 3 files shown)


oem/ibm/libpldmresponder/utils.hpp
```diff
@@ -29,13 +28,11 @@ struct CustomFD
     CustomFD(CustomFD&&) = delete;
     CustomFD& operator=(CustomFD&&) = delete;
 
-    CustomFD(int fd, bool closeOnOutScope = true) :
-        fd(fd), closeOnOutScope(closeOnOutScope)
-    {}
+    CustomFD(int fd) : fd(fd) {}
```
Pavithrab7, line 31: what happens to the fd in async case? do we not need the outOfScope check?

oem/ibm/libpldmresponder/file_io.hpp
```diff
@@ -122,13 +122,23 @@ class DMA
...
      */
-    int getNewXdmaFd()
+    int getNewXdmaFd(bool isBlckSocket = true)
     {
         try
         {
-            xdmaFd = open(xdmaDev, O_RDWR | O_NONBLOCK);
+            if (isBlckSocket)
+            {
+                xdmaFd = open(xdmaDev, O_RDWR | O_NONBLOCK);
```
Pavithrab7, line 136: here you are checking if its a blockSocket and then opening it in O_NONBLOCK mode.

Pavithrab7 (changes requested):

oem/ibm/libpldmresponder/file_io.hpp
Patel-Kamalkumar: yea correct, will change

oem/ibm/libpldmresponder/utils.hpp
Patel-Kamalkumar: this is the extra CustomFD class in async case which was not used but in sync case it's required to close when object out of scope.
in async case, DMA class handles both file descriptor

manojkiraneda (approved):

jaypadath (approved):

Pavithrab7 (approved):

PR #572, opened 2024-07-12
oem-ibm: PLDM Power down effecter and APR support

vkaverap:
This commit adds a single OEM state effecter with a OEM entity type and with states [7, 9, 10] to perform 3 custom actions ReIPL, PowerDown and ArmAPR.

The defined states are
PLDM_POWER_CYCLE_OFF_SOFT_GRACEFUL= 7,
PLDM_POWER_OFF_SOFT_GRACEFUL= 9 and
PLDM_POWER_OFF_HARD_GRACEFUL= 10

When PHYP toggles the PLDM_POWER_CYCLE_OFF_SOFT_GRACEFUL we set the state manager property RequestedHostTransition to ForceWarmReboot instead of Reboot, this will directly reboot the host firmware without notification and chassis power will remain on throughout the reboot, instead of doing a graceful reboot.

### DIFF
--- a/libpldmresponder/oem_handler.hpp
+++ b/libpldmresponder/oem_handler.hpp
@@ -22,6 +22,7 @@ class Handler : public CmdHandler
      *
      *  @param[in] entityType - entity type corresponding to the sensor
      *  @param[in] entityInstance - entity instance number
+     *  @param[in] entityContainerID - container id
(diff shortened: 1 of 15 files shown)


manojkiraneda:
This is no longer needed for 1110, Its already merged from a different PR.

PR #407, opened 2023-05-10
Marker Lid validate for Inband updates

sagisin:
This commit adds changes to call the validate API
from phosphor-software-manager to validate the
marker lid. And further send the status of the validation to PHYP through a sensor event.

PR #584, opened 2024-08-06
1110 marker lid

sagisin:
1. Moving fetch activation property (#449)
    
    This commit adds change to move the fetching of activation
    property inside the inband code update path. As it is a no-op
    for out of band updates.
    
 2. Throw Marker Lid validate error (#446)
    
    Commit the marker lid validate errors for the
    min-mif level and access key expiration so that
    it generates an SRC when we hit the error during
    code update.

3. Return proper Success/Error code to host (#442)
    
    This commit adds change to send the right error code
    to host during the marker lid validation. When the marker
    lid validation fails we were trying to send a PLDM_ERROR
    But it was sent as a response to the file write request
    and PHYP assumed that the error code returned was for the
    write operation and did not go ahead to check the sensor
    event sent to indicate the entitlment fail during marker
    lid validation.
    
    Defect: PE00LNC2

4. Marker Lid validate for Inband updates (#407)
    
    This commit adds changes to call the validate API
    from phosphor-software-manager to validate the
    marker lid. And further send the status of the validation
    to PHYP through a sensor event.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_lid.hpp
+++ b/oem/ibm/libpldmresponder/file_io_type_lid.hpp
@@ -109,7 +109,7 @@ class LidHandler : public FileHandler
 
     void validateMarkerLid(oem_platform::Handler* oemPlatformHandler)
     {
-        if (oemPlatformHandler != nullptr)
+        if (oemPlatformHandler)
         {
             pldm::responder::oem_ibm_platform::Handler* oemIbmPlatformHandler =
                 dynamic_cast<pldm::responder::oem_ibm_platform::Handler*>(
@@ -128,22 +128,22 @@ class LidHandler : public FileHandler
                     "/xyz/openbmc_project/software",
                     "xyz.openbmc_project.Software.LID", "Validate");
                 method.append(markerLidDirPath.c_str());
-                bus.call(method);
+                bus.call(method, dbusTimeout);
             }
             catch (const sdbusplus::exception::exception& e)
             {
-                if (strcmp(e.name(), accessKeyExpired) == 0)
+                if (std::string(e.name()) == accessKeyExpired)
                 {
                     phosphor::logging::commit<AccessKeyExpired>();
                     validateStatus = ENTITLEMENT_FAIL;
                 }
-                else if (strcmp(e.name(), incompatibleErr) == 0)
+                else if (std::string(e.name()) == incompatibleErr)
                 {
                     phosphor::logging::commit<IncompatibleErr>();
                     validateStatus = MIN_MIF_FAIL;
                 }
-                std::cerr << "Marker lid validate error, "
-                          << "ERROR=" << e.what() << std::endl;
+                error("Marker lid validate error, ERROR={ERR_EXCEP}",
+                      "ERR_EXCEP", e.what());
             }
             oemIbmPlatformHandler->sendStateSensorEvent(
                 sensorId, PLDM_STATE_SENSOR_STATE, 0, validateStatus, VALID);
@@ -278,7 +278,7 @@ class LidHandler : public FileHandler
     {
         int rc = PLDM_SUCCESS;
         bool codeUpdateInProgress = false;
-        if (oemPlatformHandler != nullptr)
+        if (oemPlatformHandler)
         {
             pldm::responder::oem_ibm_platform::Handler* oemIbmPlatformHandler =
                 dynamic_cast<pldm::responder::oem_ibm_platform::Handler*>(
@@ -354,7 +354,7 @@ class LidHandler : public FileHandler
         if (lidType == PLDM_FILE_TYPE_LID_MARKER)
         {
             markerLIDremainingSize -= length;
-            if (markerLIDremainingSize == 0)
+            if (!markerLIDremainingSize)
             {
                 rc = processCodeUpdateLid(lidPath);
                 if (rc == PLDM_SUCCESS)
--- a/oem/ibm/libpldmresponder/file_io_type_lid.hpp
+++ b/oem/ibm/libpldmresponder/file_io_type_lid.hpp
@@ -1,7 +1,10 @@
 #pragma once
 
 #include "file_io_by_type.hpp"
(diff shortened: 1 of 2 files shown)


oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -94,6 +107,49 @@ class LidHandler : public FileHandler
         return true;
     }
 
+    void validateMarkerLid(oem_platform::Handler* oemPlatformHandler)
+    {
+        if (oemPlatformHandler != nullptr)
```
manojkiraneda, line 112: if(oemPlatformHandler)

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -94,6 +107,49 @@ class LidHandler : public FileHandler
...
+            uint8_t validateStatus = VALID;
+            try
+            {
+                auto& bus = pldm::utils::DBusHandler::getBus();
+                auto method = bus.new_method_call(
+                    "xyz.openbmc_project.Software.BMC.Updater",
+                    "/xyz/openbmc_project/software",
+                    "xyz.openbmc_project.Software.LID", "Validate");
+                method.append(markerLidDirPath.c_str());
+                bus.call(method);
```
manojkiraneda, line 131: do we need an blocking cal for 25 seconds ? or 5 seconds is enough ?

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -94,6 +107,49 @@ class LidHandler : public FileHandler
...
+            }
+            catch (const sdbusplus::exception::exception& e)
+            {
+                if (strcmp(e.name(), accessKeyExpired) == 0)
```
manojkiraneda, line 135: strcmp() is a c style function , prefer c++ style

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -94,6 +107,49 @@ class LidHandler : public FileHandler
...
+                    phosphor::logging::commit<AccessKeyExpired>();
+                    validateStatus = ENTITLEMENT_FAIL;
+                }
+                else if (strcmp(e.name(), incompatibleErr) == 0)
+                {
+                    phosphor::logging::commit<IncompatibleErr>();
+                    validateStatus = MIN_MIF_FAIL;
+                }
+                std::cerr << "Marker lid validate error, "
+                          << "ERROR=" << e.what() << std::endl;
```
manojkiraneda, line 146: lg2

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -94,6 +107,49 @@ class LidHandler : public FileHandler
...
+                bus.call(method);
+            }
+            catch (const sdbusplus::exception::exception& e)
+            {
+                if (strcmp(e.name(), accessKeyExpired) == 0)
+                {
+                    phosphor::logging::commit<AccessKeyExpired>();
+                    validateStatus = ENTITLEMENT_FAIL;
+                }
+                else if (strcmp(e.name(), incompatibleErr) == 0)
```
manojkiraneda, line 140: same here

oem/ibm/libpldmresponder/file_io_type_lid.hpp
```diff
@@ -138,8 +194,21 @@ class LidHandler : public FileHandler
             return;
         }
         close(fd);
+
         transferFileData(lidPath, false, offset, length, address,
                          sharedAIORespDataobj, event);
+        if (lidType == PLDM_FILE_TYPE_LID_MARKER)
+        {
+            markerLIDremainingSize -= length;
+            if (markerLIDremainingSize == 0)
```
manojkiraneda, line 203: if (!markerLIDremainingSize)

oem/ibm/libpldmresponder/inband_code_update.cpp
```diff
@@ -224,13 +224,43 @@ void CodeUpdate::setVersions()
...
+                            break;
+                        }
+                    }
+                }
+                catch (const std::exception& e)
+                {
+                    error(
+                        "Code Update: Error getting Activation property, PATH={PATH} INTERFACE={INTF} PROPERTY = Activation ERROR={ERR_EXCEP}",
+                        "PATH", imageObjPath, "INTF", imageInterface,
+                        "ERR_EXCEP", e.what());
```
manojkiraneda, line 254: friendly log & e.what() should be e

manojkiraneda (changes requested):

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: done

oem/ibm/libpldmresponder/file_io_type_lid.hpp
sagisin: done

oem/ibm/libpldmresponder/inband_code_update.cpp

manojkiraneda (approved):

PR #586, opened 2024-08-06
Move the oemUtilsHandler declaration out of oem-ibm

Pavithrab7:
The oemUtilsHandler was declared in oem-ibm scope and accessed outside oem-ibm scope which would cause a problem when pldm was compiled with oem-ibm disabled.

### DIFF
--- a/pldmd/pldmd.cpp
+++ b/pldmd/pldmd.cpp
@@ -241,13 +241,13 @@ int main(int argc, char** argv)
     platformConfigHandler =
         std::make_unique<platform_config::Handler>(PDR_JSONS_DIR);
     std::unique_ptr<oem_fru::Handler> oemFruHandler{};
+    std::unique_ptr<oem_utils::Handler> oemUtilsHandler{};
 
 #ifdef OEM_IBM
     respInterface.responseObj =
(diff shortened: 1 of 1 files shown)


manojkiraneda (approved):

pldmd/pldmd.cpp
```diff
@@ -241,13 +241,13 @@ int main(int argc, char** argv)
     platformConfigHandler =
         std::make_unique<platform_config::Handler>(PDR_JSONS_DIR);
     std::unique_ptr<oem_fru::Handler> oemFruHandler{};
+    std::unique_ptr<oem_utils::Handler> oemUtilsHandler{};
```
Patel-Kamalkumar, line 244: please make it as oemUtilsHandler=nullptr

Pavithrab7: Both type of declaration initializes the oemUtilsHandler to nullptr itself, one is explicitly assigns the nullptr, another one uses brace initialization ({}), which is equivalent to default initialization for std::unique_ptr.

Is my understanding correct?

Patel-Kamalkumar: yup, just checked both are same, no need to change.

Patel-Kamalkumar (approved):

PR #599, opened 2024-08-20
Fixing race condition during file transfer

Patel-Kamalkumar:
there is a race condition occurs between performing post operation and
sending response to Host. When host sends pels then bmc takes some time
to store the pels and meanwhile host is timed-out so it will try again
to transfer the pels and suddenly BMC finishing the storing and sending
failure status to host which causing the problem.

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_by_type.cpp
+++ b/oem/ibm/libpldmresponder/file_io_by_type.cpp
@@ -160,15 +160,16 @@ void FileHandler::transferFileData(int fd, bool upstream, uint32_t offset,
         }
         if (static_cast<int>(data.length) == rc)
         {
-            wInterface->setTransferStatus(true);
-            dmaResponseToRemoteTerminus(sharedAIORespDataobj, PLDM_SUCCESS,
-                                        origLength);
             if (sharedAIORespDataobj.functionPtr != nullptr)
             {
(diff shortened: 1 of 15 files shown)


oem/ibm/libpldmresponder/file_io_by_type.cpp
```diff
@@ -39,7 +39,8 @@ void FileHandler::dmaResponseToRemoteTerminus(
     const SharedAIORespData& sharedAIORespDataobj,
     const pldm_completion_codes rStatus, uint32_t length)
 {
-    Response response(sizeof(pldm_msg_hdr) + sharedAIORespDataobj.command, 0);
+    Response response(
```
Pavithrab7, line 42: once this PR goes in I guess we can remove this change - https://github.com/ibm-openbmc/pldm/pull/598

Pavithrab7 (changes requested):
@Patel-Kamalkumar do you want to give this patch to Stephanie as well to see if we hit that issue again?

oem/ibm/libpldmresponder/file_io_by_type.cpp
Patel-Kamalkumar: done

manojkiraneda (approved):

Pavithrab7 (approved):

PR #600, opened 2024-08-23
pdr: Log a PEL only when Host is Running

Pavithrab7:
When there is a PDR exchange happening with host and we trigger a poweroff we were logging a pel BD8D6026 when the PDR exchange fails. This change is needed check if the host is up or not and then log an error only if host is up.

### DIFF
--- a/host-bmc/host_pdr_handler.cpp
+++ b/host-bmc/host_pdr_handler.cpp
@@ -113,7 +113,7 @@ HostPDRHandler::HostPDRHandler(
     oemUtilsHandler(oemUtilsHandler)
 {
     isHostOff = false;
-    isHostTransitioningToOff = false;
+    isHostRunning = false;
     mergedHostParents = false;
     hostOffMatch = std::make_unique<sdbusplus::bus::match_t>(
         pldm::utils::DBusHandler::getBus(),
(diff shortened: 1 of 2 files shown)


host-bmc/host_pdr_handler.cpp
```diff
@@ -149,7 +149,7 @@ HostPDRHandler::HostPDRHandler(
                 this->stateSensorPDRs.clear();
                 fruRecordSetPDRs.clear();
                 isHostOff = true;
-                isHostTransitioningToOff = false;
+                isHostRunning = false;
```
ArchanaKakani, line 152: isHostRunning should also be reset when host goes to any state apart from Running.

Pavithrab7: yes, made the changes.

ArchanaKakani (approved):

host-bmc/host_pdr_handler.cpp
```diff
@@ -162,18 +162,16 @@ HostPDRHandler::HostPDRHandler(
...
-                "xyz.openbmc_project.State.Host.HostState.TransitioningToOff")
-            {
-                isHostTransitioningToOff = true;
-            }
             else if (propVal ==
                      "xyz.openbmc_project.State.Host.HostState.Running")
             {
-                isHostTransitioningToOff = false;
+                isHostRunning = true;
                 isHostOff = false;
```
manojkiraneda, line 169: can't we just use the `isHostOff` variable instead ? why to create a new variable ? , since isHostOff has inverse relation to isHostRunning, we can directly use `!isHostOff` where ever we were using `isHostRunning` ?

manojkiraneda: looked again , it looks risky now. lets not do it.

manojkiraneda (approved):

PR #604, opened 2024-09-06
CI: Adding format change

riyadixitagra:
Due to new change in openbmc-build-scripts for the 1110 branch to clang-18, a lot of files were failing in CI format. Adding the changed format files in correspondance to CI update.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2024-04-26T19:13:37Z",
+  "generated_at": "2024-09-06T12:22:16Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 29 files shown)


.secrets.baseline
```diff
@@ -87,24 +87,6 @@
         "verified_result": null
       }
     ],
-    "oem/ibm/pfw-sms-utils/pfw_sms_menu.cpp": [
```
Pavithrab7, line 90: why was this removed?

Pavithrab7 (changes requested):

.secrets.baseline
manojkiraneda: +1

manojkiraneda (changes requested):

.secrets.baseline
jaypadath: Rather .secrets.baseline is a generated file ? For this commit to I think that file is not needed.

riyadixitagra: okk

riyadixitagra: But this commit has no change other than what running the local CI caused, so I added it. Please let me know how to handle this for CI to pass?

jaypadath: Riya, you can ignore it as that deleted pfw file is not yet present in 1110-ghe and it will only be checked in with my SMS changes. Secrets file was just added for CI to pass automatically. So I am good here.

jaypadath (approved):

.secrets.baseline
manojkiraneda: cool, sounds good then.

manojkiraneda (approved):

PR #606, opened 2024-09-10
libpldmresponder: fix an eternal hang in requesting bus name

Pavithrab7:
If oem-ibm is disabled, but system-specific-bios-json option is enabled pldmd fails to start since its timing out waiting to request the bus name. This fix also breaks the dependency between the above two meson options.


Pavithrab7:
Upstream comimit- https://gerrit.openbmc.org/c/openbmc/pldm/+/72632

manojkiraneda (approved):

PR #608, opened 2024-09-13
Remove check for IBM cable card

vkaverap:
Removing the check for IBM cable card because this will be verified in VPD and there is no need to have an additional check inside PLDM

Tested by : performing CM on IBM cable cards and industry standard adapters

### DIFF
--- a/libpldmresponder/fru.cpp
+++ b/libpldmresponder/fru.cpp
@@ -921,7 +921,7 @@ void FruImpl::processFruPresenceChange(const DbusChangedProps& chProperties,
 #ifdef OEM_IBM
     if (fruInterface == "xyz.openbmc_project.Inventory.Item.PCIeDevice")
     {
-        if (!pldm::responder::utils::checkIfIBMCableCard(fruObjPath))
+        if (!pldm::responder::utils::checkFruPresence(fruObjPath.c_str()))
         {
             return;
         }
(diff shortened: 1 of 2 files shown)


libpldmresponder/fru.cpp
```diff
@@ -921,7 +921,9 @@ void FruImpl::processFruPresenceChange(const DbusChangedProps& chProperties,
 #ifdef OEM_IBM
     if (fruInterface == "xyz.openbmc_project.Inventory.Item.PCIeDevice")
     {
-        if (!pldm::responder::utils::checkIfIBMCableCard(fruObjPath))
+        bool isPresent =
+            pldm::responder::utils::checkFruPresence(fruObjPath.c_str());
+        if (!isPresent)
```
Pavithrab7, line 926: can we change it to if(!pldm::responder::utils::checkFruPresence(fruObjPath.c_str())  to reduce the bool expression?

vkaverap: ok how do we check for OEM_IBM?

Pavithrab7: we do not need isPresent param right to check for OEM_IBM. we can skip creating that param is what I suggested.

Pavithrab7 (changes requested):

ArchanaKakani (commented):
With this change, PLDM will create the PDRs only if the FRU is present.
But in the case where Dbus query fails, we are returning true. Is that the expected behaviour?

libpldmresponder/fru.cpp
vkaverap: ok done

manojkiraneda (approved):

jaypadath (approved):

PR #614, opened 2024-09-25
pdr: Add some FFDC for debugging

Pavithrab7:
There is a intermittent issue that happens when we fail to send a repo change event to HB after a PDR exchange and this scenario is a one-off and is difficult to recreate manually. Adding these traces will help to improve the FFDC and debugging.

The commit also made some changes to the code which was wrong in 1110 when compared to the 1060 branch.

Change-Id: I00f2226fb6a4f6a3cc2f49cd5aaa0d03230a8602

### DIFF
--- a/common/utils.cpp
+++ b/common/utils.cpp
@@ -1,7 +1,10 @@
 #include "utils.hpp"
 
+#include "libpldm/platform.h"
+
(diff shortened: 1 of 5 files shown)


ArchanaKakani (commented):
Do we need to ModifyCounter value ??

PR #621, opened 2024-10-10
oem-ibm: Handle the json parse exception correctly

jaypadath:
This commit handles the CoD license json parsing exception by correctly catching it and returs the error to the caller. Also made the change to avoid the removal of license json file whenever shared by the host.

Tested:
Done the sanity testing to verify that json file was kept without removal.

### DIFF
--- a/libpldmresponder/bios_config.cpp
+++ b/libpldmresponder/bios_config.cpp
@@ -128,11 +128,20 @@ void BIOSConfig::initBIOSAttributes(const std::string& systemType,
 
     buildTables();
 
+    info("Register listenPendingAttributes");
+    listenPendingAttributes();
+
     if (pendingAttributes.size())
     {
         try
         {
-            info("Setting Pending Attributes");
+            for (const auto& pair : pendingAttributes)
+            {
+                info("Pending Attribute is {A}", "A", pair.first);
+            }
+
+            info("Setting Pending Attributes {NUMBER}", "NUMBER",
+                 (int)pendingAttributes.size());
             auto method = bus.new_method_call(service.c_str(), biosObjPath,
                                               dbusProperties, "Set");
             method.append(biosInterface, "PendingAttributes",
@@ -145,8 +154,6 @@ void BIOSConfig::initBIOSAttributes(const std::string& systemType,
                   "ERROR", e);
         }
     }
-    info("Register listenPendingAttributes");
-    listenPendingAttributes();
     if (registerService)
     {
         requestPLDMServiceName();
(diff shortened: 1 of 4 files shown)


oem/ibm/libpldmresponder/file_io_type_lic.cpp
```diff
@@ -33,13 +33,17 @@ int LicenseHandler::updateBinFileAndLicObjs(const fs::path& newLicJsonFilePath)
     int rc = PLDM_SUCCESS;
     fs::path newLicFilePath(fs::path(licFilePath) / newLicenseFile);
     std::ifstream jsonFileNew(newLicJsonFilePath);
+    Json dataNew;
 
-    auto dataNew = Json::parse(jsonFileNew, nullptr, false);
-    if (dataNew.is_discarded())
+    try
+    {
+        dataNew = Json::parse(jsonFileNew);
```
manojkiraneda, line 40: define dataNew here itself? auto dataNew

oem/ibm/libpldmresponder/file_io_type_lic.cpp
```diff
@@ -33,13 +33,17 @@ int LicenseHandler::updateBinFileAndLicObjs(const fs::path& newLicJsonFilePath)
...
+    }
+    catch (const Json::parse_error& e)
     {
         error("Failed to parse the new license json file '{NEW_LIC_JSON}'",
               "NEW_LIC_JSON", newLicJsonFilePath);
-        throw InternalFailure();
+        return PLDM_ERROR;
```
manojkiraneda, line 46: return PLDM_ERROR_INVALID_DATA ?

oem/ibm/libpldmresponder/file_io_type_lic.cpp
jaypadath: We need to use this dataNew after try-catch, so can't use auto

Pavithrab7, line 40: Jay, no need of other two parameters?  (nullptr and false)

Pavithrab7 (changes requested):

oem/ibm/libpldmresponder/file_io_type_lic.cpp
jaypadath: Not needed as the they default args with values as nullptr and true

oem/ibm/libpldmresponder/file_io_type_lic.cpp
jaypadath: Done

jaypadath:
jenkins run tests please

Pavithrab7 (approved):

libpldmresponder/bios_config.cpp
```diff
@@ -128,11 +128,15 @@ void BIOSConfig::initBIOSAttributes(const std::string& systemType,
...
+    info("Register listenPendingAttributes");
+    listenPendingAttributes();
+
     if (pendingAttributes.size())
     {
         try
         {
-            info("Setting Pending Attributes");
+            info("Setting Pending Attributes {NUMBER}", "NUMBER",
+                 (int)pendingAttributes.size());
```
manojkiraneda, line 139: can't we print the attribute names here ? size might not give much clues if we hit this path.

manojkiraneda (approved):

PR #623, opened 2024-10-15
configs: Add PCIe slot power states for Bonnell

spinler:
Add the entries to Bonnell's 11.json so that the host can send down the PCIE slot power states for the four PCIe slots.  These are the same as on Rainier and Everest.

Tested:
The power state in phosphor-inventory-manager is accurately reflected for a plugged in card.

Change-Id: Ic04e6d11750add7656bc7ac40135e2bfa4b8490e

### DIFF
--- a/oem/ibm/configurations/pdr/ibm,bonnell/11.json
+++ b/oem/ibm/configurations/pdr/ibm,bonnell/11.json
@@ -402,6 +402,98 @@
                             }
                         }
                     ]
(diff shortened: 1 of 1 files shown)


spinler:
This is a straight cherry pick of https://gerrit.openbmc.org/c/openbmc/pldm/+/75305

manojkiraneda (approved):

PR #627, opened 2024-11-04
pldm: Fix the incorrect processor core count

jaypadath:
This commit fixes the issue of calculating the processor core count incorrectly.

Fixes: STG Defect 658536

Tested By:
Tested the changes in a Blue ridge system and verified that, the processor core count is coming up correctly under each of the processors.

### DIFF
--- a/oem/ibm/libpldmresponder/utils.cpp
+++ b/oem/ibm/libpldmresponder/utils.cpp
@@ -449,7 +449,7 @@ using namespace pldm::utils;
 int pldm::responder::oem_ibm_utils::Handler::setCoreCount(
     const EntityAssociations& Associations, const EntityMaps entityMaps)
 {
-    int coreCountRef = 0;
+    int coreCount = 0;
     // get the CPU pldm entities
     for (const auto& entries : Associations)
     {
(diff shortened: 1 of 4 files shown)


oem/ibm/libpldmresponder/utils.cpp
```diff
@@ -449,15 +449,16 @@ using namespace pldm::utils;
 int pldm::responder::oem_ibm_utils::Handler::setCoreCount(
     const EntityAssociations& Associations, const EntityMaps entityMaps)
 {
-    int coreCountRef = 0;
+    int coreCount = 0;
```
manojkiraneda, line 452: we dont need this anymore ?

oem/ibm/libpldmresponder/utils.cpp
```diff
@@ -449,15 +449,16 @@ using namespace pldm::utils;
...
     // get the CPU pldm entities
     for (const auto& entries : Associations)
     {
         auto parent = pldm_entity_extract(entries[0]);
         // entries[0] would be the parent in the entity association map
         if (parent.entity_type == PLDM_ENTITY_PROC)
         {
-            int& coreCount = coreCountRef;
+            // Need to reset the count for each of the processors
+            coreCount = 0;
```
manojkiraneda, line 461: just make this `int coreCount =0` ?

manojkiraneda (changes requested):
This seems like a miss from 1020 migration.  Changes looks good to me. minor nitpick.

oem/ibm/libpldmresponder/utils.cpp
jaypadath: We needed like this as we need to return the core count for unit test code.

manojkiraneda (approved):

PR #629, opened 2024-11-08
pdr:Fix multiple adapter issue

ArchanaKakani:


libpldmresponder/fru.cpp
```diff
@@ -918,33 +926,40 @@ void FruImpl::processFruPresenceChange(const DbusChangedProps& chProperties,
...
-#endif
 
     // if(fruInterface != "xyz.openbmc_project.Inventory.Item.PCIeDevice")
     {
         if (newPropVal)
         {
+            info("Presence changed to true");
             buildIndividualFRU(fruInterface, fruObjPath);
-            for (auto portObject : portObjects)
+            if((fruInterface == "xyz.openbmc_project.Inventory.Item.PCIeDevice") && (checkIfIBMCableCard(fruObjPath)))
```
Pavithrab7, line 936: should this not be` !(checkIfIBMCableCard(fruObjPath)` ?

libpldmresponder/fru.cpp
```diff
@@ -544,6 +544,13 @@ void FruImpl::buildIndividualFRU(const std::string& fruInterface,
     pldm_entity parentEntity{};
     static uint32_t last_bmc_record_handle = 0;
     uint32_t newRecordHdl{};
+
+    if(fruInterface == "xyz.openbmc_project.Inventory.Item.PCIeDevice")
+    {
+        info("Removing old PDRs {PATH}", "PATH", fruObjectPath);
+        removeIndividualFRU(fruObjectPath);
```
Pavithrab7, line 551: if we remove this will we not lose the LED state values corresponding the card and all? (not sure though)

libpldmresponder/fru.cpp
ArchanaKakani: checkIfIBMCableCard function returns true if the adapter is IBM Cable card. And we expect ports under cable card. 
https://github.com/ibm-openbmc/pldm/blob/391c4012a85fb85d5b9d9a37acd63f049d9616fc/oem/ibm/libpldmresponder/utils.cpp#L427

vkaverap (approved):

PR #636, opened 2024-11-22
CM Fix

ArchanaKakani:


Pavithrab7 (approved):

manojkiraneda (approved):

libpldmresponder/fru.cpp
```diff
@@ -915,40 +928,61 @@ void FruImpl::processFruPresenceChange(const DbusChangedProps& chProperties,
...
+                {
+                    buildIndividualFRU(portInterface, portObject);
+                }
+                return;
+            }
+            else
+            {
+                // PDRs realted to industry standard card are not
+                // rebuilt.
+                info("Skipping building fru for {PATH}", "PATH", fruObjPath);
```
manojkiraneda, line 960: wouldn't these flood the journal ?

ArchanaKakani: Trace will be printed during slot disable operation triggered after poweron. I would like to keep this trace for few days for debug. We can remove it after a month or two

PR #641, opened 2024-12-09
libresponder: Fixing issues in CM path (#636)

ArchanaKakani:
There is multiple issues faced during the CM. This commit
    fixes below issues
    1. Port PDRs not getting generated after adding cable card concurrently
    2. Multiple adapter were seen under single I/O Slot
    3. No adapter PDRs for the cards under NVMe Slots (Drive Backplane)

    When resolving issues, the design is that the Adapter PDR will always
    remain in the PDR repository, regardless of whether the card is
    physically present or absent.

    Tested:
    Using busctl command


libpldmresponder/fru.cpp
```diff
@@ -922,37 +941,65 @@ void FruImpl::processFruPresenceChange(const DbusChangedProps& chProperties,
...
-        "xyz.openbmc_project.Inventory.Item.Connector";
 
-    if (oemUtilsHandler)
+    if (newPropVal)
     {
+        info("Presence changed to true for {PATH}", "PATH", fruObjPath);
         if (fruInterface == "xyz.openbmc_project.Inventory.Item.PCIeDevice")
         {
-            if (!oemUtilsHandler->checkFruPresence(fruObjPath.c_str()))
+#ifdef OEM_IBM
```
Pavithrab7, line 950: can we remove this if def's and add if(oemUtilsHandler) ?

Pavithrab7 (approved):

PR #642, opened 2024-12-09
pldm: LI 1TM IBM OEM FRU Names Support

jaypadath:
This commit supports the addition of IBM OEM FRU Names. Pretty name of the inventory item is updated by extracting the Entity Auxiliary Names PDR passed from host.

Tested By:
Verified that the decode of Entity Auxiliary Names PDR results in the intended pretty name string.

### DIFF
--- a/host-bmc/dbus/custom_dbus.cpp
+++ b/host-bmc/dbus/custom_dbus.cpp
@@ -114,8 +114,9 @@ void CustomDBus::setLinkReset(
     link.at(path)->linkReset(value);
 }
 
-void CustomDBus::updateItemPresentStatus(const std::string& path,
-                                         bool isPresent)
+void CustomDBus::updateInventoryItemProperties(
+    const std::string& path, bool isPresent,
+    const std::optional<std::string>& prettyName)
 {
     if (presentStatus.find(path) == presentStatus.end())
     {
@@ -127,14 +128,28 @@ void CustomDBus::updateItemPresentStatus(const std::string& path,
         // Hardcode the present dbus property to true
         presentStatus.at(path)->present(true);
 
-        // Set the pretty name dbus property to the filename
-        // form the dbus path object
-        presentStatus.at(path)->prettyName(ObjectPath.filename());
+        if (prettyName.has_value() && !prettyName.value().empty())
+        {
+            // Set the pretty name dbus property
+            presentStatus.at(path)->prettyName(prettyName.value());
+        }
+        else
+        {
+            // Set the pretty name dbus property to the filename
+            // form the dbus path object
+            presentStatus.at(path)->prettyName(ObjectPath.filename());
+        }
     }
     else
     {
         // object is already created
         presentStatus.at(path)->present(isPresent);
+
+        if (prettyName.has_value() && !prettyName.value().empty())
+        {
+            // Set the pretty name dbus property
+            presentStatus.at(path)->prettyName(prettyName.value());
+        }
     }
 }
 
--- a/host-bmc/host_pdr_handler.cpp
+++ b/host-bmc/host_pdr_handler.cpp
@@ -145,6 +145,7 @@ HostPDRHandler::HostPDRHandler(
                     this->mergedHostParents = false;
                     this->stateSensorPDRs.clear();
                     fruRecordSetPDRs.clear();
+                    entityAuxNamesPDRs.clear();
                     isHostOff = true;
                     isHostRunning = false;
                     this->sensorIndex = stateSensorPDRs.begin();
@@ -415,16 +416,8 @@ void HostPDRHandler::mergeEntityAssociations(
 
         else
         {
-            uint16_t terminus_handle = 0;
-            for (const auto& [terminusHandle, terminusInfo] : tlPDRInfo)
-            {
-                if (std::get<0>(terminusInfo) == terminusID &&
-                    std::get<1>(terminusInfo) == mctp_eid &&
-                    std::get<2>(terminusInfo))
-                {
-                    terminus_handle = terminusHandle;
-                }
-            }
+            uint16_t terminus_handle = getRemoteTerminusHandle();
+
             if ((isHostUp() || (terminus_handle & 0x8000)) &&
                 oemPlatformHandler)
             {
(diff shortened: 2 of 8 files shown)


host-bmc/dbus/custom_dbus.cpp
```diff
@@ -92,8 +92,9 @@ void CustomDBus::implementCableInterface(const std::string& path)
     }
 }
 
-void CustomDBus::updateItemPresentStatus(const std::string& path,
-                                         bool isPresent)
+void CustomDBus::updateInventoryItemProperties(const std::string& path,
+                                               bool isPresent,
+                                               const std::string& prettyName)
```
manojkiraneda, line 97: make this an std::optional<std::string>& instead

host-bmc/dbus/custom_dbus.cpp
```diff
@@ -105,9 +106,17 @@ void CustomDBus::updateItemPresentStatus(const std::string& path,
         // Hardcode the present dbus property to true
         presentStatus.at(path)->present(true);
 
-        // Set the pretty name dbus property to the filename
-        // form the dbus path object
-        presentStatus.at(path)->prettyName(ObjectPath.filename());
+        if (prettyName.empty())
```
manojkiraneda, line 109: if it is an std::optional string, then you can just check if the value is present  - rather than checking for an string which is empty(), then use it ..otherwise use the ObjectPath.filename().

host-bmc/host_pdr_handler.cpp
```diff
@@ -760,6 +761,12 @@ void HostPDRHandler::processHostPDRs(mctp_eid_t /*eid*/,
                     updateContainerId<pldm_numeric_effecter_value_pdr>(
                         entityTree, pdr);
                 }
+                else if (pdrHdr->type == PLDM_ENTITY_AUXILIARY_NAMES_PDR)
+                {
+                    pdrTerminusHandle = 0xFFFF;
```
manojkiraneda, line 766: why was this needed ?

host-bmc/host_pdr_handler.cpp
```diff
@@ -1838,6 +1877,92 @@ void HostPDRHandler::setOperationStatus()
     }
 }
 
+std::shared_ptr<EntityAuxiliaryNames>
+    HostPDRHandler::parseEntityAuxNamesPDR(std::vector<uint8_t>& pdrData)
```
manojkiraneda, line 1881: this is code duplication. I think its better to use the existing code by moving the parse function into the utils.cpp

manojkiraneda (changes requested):

host-bmc/dbus/custom_dbus.cpp
jaypadath: Done

host-bmc/host_pdr_handler.cpp
jaypadath: There is no specific terminus handle for PLDM_ENTITY_AUXILIARY_NAMES_PDR. So I put it as a default value instead of 0.

host-bmc/host_pdr_handler.cpp
jaypadath: Yes this API almost matches with https://github.com/openbmc/pldm/blob/master/platform-mc/terminus.cpp#L307 except that in our use case we are not trimming the name. As of now I am moving this to a util API.

host-bmc/dbus/custom_dbus.cpp
jaypadath: Done

host-bmc/utils.cpp
```diff
@@ -265,6 +265,65 @@ EntityMaps parseEntityMap(const fs::path& filePath)
...
+            std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}
+                .to_bytes(u16NameString);
+        nameStrings.emplace_back(std::make_pair(nameLanguageTag, nameString));
+    }
+
+    EntityKey key{decodedPdr->container.entity_type,
+                  decodedPdr->container.entity_instance_num,
+                  decodedPdr->container.entity_container_id};
+
+    return std::make_shared<EntityAuxiliaryNames>(key, nameStrings);
```
ArchanaKakani, line 324: is there any specific reason to use shared ptr here? Can't we avoid them?

host-bmc/host_pdr_handler.cpp
```diff
@@ -1840,6 +1879,53 @@ void HostPDRHandler::setOperationStatus()
...
+
+    auto it = std::find_if(
+        entityAuxiliaryNamesList.begin(), entityAuxiliaryNamesList.end(),
+        [entity](
+            const std::shared_ptr<EntityAuxiliaryNames>& entityAuxiliaryNames) {
+        const auto& [key, entityNames] = *entityAuxiliaryNames;
+        return (entityAuxiliaryNames && key.type == entity.entity_type &&
+                key.instanceIdx == entity.entity_instance_num &&
+                key.containerId == entity.entity_container_id &&
+                entityNames.size());
```
ArchanaKakani, line 1899: this will become simple if we use map for entityAuxiliaryNamesList.

host-bmc/host_pdr_handler.hpp
```diff
@@ -452,6 +471,13 @@ class HostPDRHandler
     PDRList stateSensorPDRs;
     PDRList fruRecordSetPDRs{};
 
+    /** @brief Entity Auxiliary Name PDRs list */
+    PDRList entityAuxNamesPDRs{};
+
+    /** @brief Entity Auxiliary Names list */
+    std::vector<std::shared_ptr<pldm::hostbmc::utils::EntityAuxiliaryNames>>
```
ArchanaKakani, line 478: I think changing the datatype of "entityAuxiliaryNamesList" to map will simplify things and avoid unnecessary calculations. Do you think otherwise?

host-bmc/utils.hpp
```diff
@@ -22,6 +24,34 @@ namespace hostbmc
 namespace utils
 {
 
+using ContainerID = uint16_t;
```
ArchanaKakani, line 27: Can we get rid of them as those are not adding much value??

ArchanaKakani (changes requested):

jaypadath:
jenkins run tests please

jaypadath:
jenkins run tests please

host-bmc/utils.cpp
jaypadath: This matches with the upstream design of parsing the entity aux names and approved by the community. I assume std::make_shared is used for catering the design of selecting the key with a combination of values.

host-bmc/host_pdr_handler.cpp
jaypadath: Same comment as above and not sure whether I will be able to upstream with a different way to parse the same PDR.

host-bmc/host_pdr_handler.hpp
jaypadath: It would be ideally same as I need to compare against each of the PDR entries to entity key parameters and there is no special calculation used here.

ArchanaKakani (approved):

manojkiraneda (approved):

PR #656, opened 2025-01-09
topology: Fix to solve multiple associations

Pavithrab7:
The PCIeDevice was associated with multiple PCIeSlots due to the error when creating the association between devices and slots. The commit fixes the association issues present.

Tested: Checked the associations and the topology page loaded fine.

Change-Id: I782f8b8b9f31ea2a8bd9ab68cbfd168eed01602f

### DIFF
--- a/host-bmc/host_pdr_handler.cpp
+++ b/host-bmc/host_pdr_handler.cpp
@@ -1728,6 +1728,10 @@ void HostPDRHandler::setFRUDynamicAssociations()
         for (const auto& [rightPath, rightEntity] : objPathMap)
         {
             uint16_t rightEntityType = rightEntity.entity_type;
+
+            std::string leftPath1 = leftPath.string() + "/";
+            std::string rightPath1 = rightPath.string() + "/";
+
             // if leftpath is same as rightPath
             // then both dbus objects are same, so
             // no need to create any associations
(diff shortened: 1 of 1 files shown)


manojkiraneda (approved):

host-bmc/host_pdr_handler.cpp
```diff
@@ -1728,15 +1728,21 @@ void HostPDRHandler::setFRUDynamicAssociations()
...
             // then both dbus objects are same, so
             // no need to create any associations
             if (leftPath == rightPath)
             {
                 continue;
             }
-            else if ((rightPath.string().find(leftPath) != std::string::npos) ||
-                     (leftPath.string().find(rightPath) != std::string::npos))
+
+            else if ((rightPath.string().find(leftPath1) !=
```
ArchanaKakani, line 1743: Can we use parent_path() to know the hierarchy?

ArchanaKakani (changes requested):

host-bmc/host_pdr_handler.cpp
Pavithrab7: I did try this and it seemed to be not working, keeping the logic as izz

ArchanaKakani (approved):

PR #675, opened 2025-02-19
libpldmresponder: Use async call in storePel

Jino26:
The storePel function used call_noreply, a synchronous D-Bus method with timeout of 5 secs, which caused it to timeout in some scenarios.

To mitigate this issue, it has been replaced with call_async method. Additionally, PLDM_SUCCESS is being returned now regardless of the method call outcome. In case of any error during the async method call, it will be logged by the callback function.

Change-Id: Ifbb80585ace516c8d4bf484c08356c3734d351fc


Jino26:
Gerrit: https://gerrit.openbmc.org/c/openbmc/pldm/+/78244

geissonator:
add to approvelist

PR #683, opened 2025-04-06
oem-ibm: Fix for the PEL parsing issue

jaypadath:
This commit contains the fix for the invalid PEL issue where the PEL send by the remote end point could not get parsed properly. Fix is to explicitly execute the unix FD sync and close after the write FD operation.

Fixes : STG Defect 678010

Change-Id: Ib81fd21e160423f103218e8839133119a5a89cc1


jaypadath:
Not needed

PR #686, opened 2025-04-08
soft-off: Return error if effecterid is not found

ArchanaKakani:
SoftOff fails to find the host effector pdr in BMC PDR repository if the softoff request arrives in the time window where HB is down and PHYP is up but PDR exchange with PHYP is not completed.

Fixed: STG Defect 689447


Pavithrab7 (approved):

manojkiraneda (approved):
please put this upstream as well.

PR #689, opened 2025-04-15
New service the clear's the stray DMA transactions

manojkiraneda:
In the current state , the DMA infrastruture written in pldm is Non-blocking IO, but for NVRAM writes/reads pldm still relies on blocking IO as AIO for NVRAM writes cause corruption especially during chassis off path where the host goes down abruptly. Due to pldm operating in blocking mode there are cases where pldm hangs in read/write system calls due to multiple reasons like :

1. Stray DMA message in mctp buffer When DMA operations in progress with PHYP and it goes down abruptly, the DMA command of PHYP could be stored inside the mctp buffer and that can be delivered to Hostboot(which does not have pcie support) leading to a pldm hang.

2. PHYP crashes while a DMA transaction is in progress xdma driver on bmc does not have a way to know if host had gone down due to crash , so the outstanding DMA operation that is currently running would be left hanging, leaving pldm to hang as well.

Added new service pldm-clear-dma.service which would send signals (SIGSTOP, SIGCONT) signals to pldm while the host is going down, so that the stray DMA operations are cancelled and pldm daemon can be recovered from the system call hangs.

### DIFF
--- a/oem/ibm/service_files/meson.build
+++ b/oem/ibm/service_files/meson.build
@@ -10,3 +10,8 @@ filesystem.copyfile(
     install: true,
     install_dir: systemd_system_unit_dir,
 )
+filesystem.copyfile(
+    'pldm-clear-dma.service',
+    install: true,
+    install_dir: systemd_system_unit_dir,
+)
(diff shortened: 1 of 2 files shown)


manojkiraneda:
fixes defect : https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=693139

ArchanaKakani (approved):

PR #705, opened 2025-06-26
oem-ibm: Add support for targeted ACF file

Jino26:
This commit introduces support for specifying a targeted ACF (Access Control File) for resource dumps. The ACF file path and name are passed to PLDM via the D-Bus property "AcfFile". This property is populated when the CreateDump method of the dump manager is invoked. The specified ACF file is then used for the resource dump, while the existing resource dump workflow remains unchanged.

### DIFF
--- a/oem/ibm/requester/dbus_to_file_handler.cpp
+++ b/oem/ibm/requester/dbus_to_file_handler.cpp
@@ -119,7 +119,8 @@ void DbusToFileHandler::reportResourceDumpFailure(const std::string_view& str)
 }
 
 void DbusToFileHandler::processNewResourceDump(
-    const std::string& vspString, const std::string& resDumpReqPass)
+    const std::string& vspString, const std::string& resDumpReqPass,
+    const std::string& acfFilePath)
 {
     try
     {
@@ -193,9 +194,9 @@ void DbusToFileHandler::processNewResourceDump(
     fileFunc(resDumpReqPass);
 
     std::string str;
-    if (!resDumpReqPass.empty())
+    if (vspString != "system")
     {
-        str = getAcfFileContent();
+        str = getAcfFileContent(acfFilePath);
     }
 
     fileFunc(str);
@@ -206,10 +207,14 @@ void DbusToFileHandler::processNewResourceDump(
     sendNewFileAvailableCmd(fileSize);
 }
 
-std::string DbusToFileHandler::getAcfFileContent()
+std::string DbusToFileHandler::getAcfFileContent(const std::string& acfPath)
 {
     std::string str;
-    static constexpr auto acfDirPath = "/etc/acf/service.acf";
+    const std::string& acfDirPath =
+        acfPath.empty() ? "/etc/acf/service.acf" : acfPath;
+
+    info("Using acf file : {ACF_FILE_PATH}", "ACF_FILE_PATH", acfDirPath);
+
     if (fs::exists(acfDirPath))
     {
         std::ifstream file;
--- a/oem/ibm/requester/dbus_to_file_handler.hpp
+++ b/oem/ibm/requester/dbus_to_file_handler.hpp
@@ -48,9 +48,11 @@ class DbusToFileHandler
     /** @brief Process the new resource dump request
      *  @param[in] vspString - vsp string
      *  @param[in] resDumpReqPass - resource dump password
+     *  @param[in] acfFilePath - ACF file path
      */
     void processNewResourceDump(const std::string& vspString,
-                                const std::string& resDumpReqPass);
+                                const std::string& resDumpReqPass,
+                                const std::string& acfFilePath = "");
 
     /** @brief Process the new CSR file available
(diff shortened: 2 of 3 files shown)


oem/ibm/requester/dbus_to_file_handler.cpp
```diff
@@ -193,9 +197,9 @@ void DbusToFileHandler::processNewResourceDump(
     fileFunc(resDumpReqPass);
 
     std::string str;
-    if (!resDumpReqPass.empty())
+    if (vspString != "system")
     {
-        str = getAcfFileContent();
+        str = getAcfFileContent(acfFilePath);
```
ArchanaKakani, line 199: Do we need to handle error case where ACF file path does not exist?

ArchanaKakani (commented):
Please Test details in commit message

oem/ibm/requester/dbus_to_file_handler.hpp
```diff
@@ -107,7 +109,7 @@ class DbusToFileHandler
     void reportResourceDumpFailure(const std::string_view& str);
 
     /** @brief method to get the acf file contents */
-    std::string getAcfFileContent();
+    std::string getAcfFileContent(const std::string acfDirPath);
```
Pavithrab7: update the doxygen with the input param

oem/ibm/requester/dbus_to_file_handler.cpp
Pavithrab7: +1, have a check or log an error?

Pavithrab7 (changes requested):

oem/ibm/requester/dbus_to_file_handler.cpp
Jino26: The `getAcfFileContent` function checks if the path/file exists.

oem/ibm/requester/dbus_to_file_handler.hpp
Jino26: Done.

Jino26:
> Please Test details in commit message

@ArchanaKakani added test details in commit.

oem/ibm/requester/dbus_to_file_handler.cpp
```diff
@@ -206,10 +210,10 @@ void DbusToFileHandler::processNewResourceDump(
     sendNewFileAvailableCmd(fileSize);
 }
 
-std::string DbusToFileHandler::getAcfFileContent()
+std::string DbusToFileHandler::getAcfFileContent(const std::string acfDirPath)
 {
     std::string str;
-    static constexpr auto acfDirPath = "/etc/acf/service.acf";
+
```
jaypadath, line 217: Won't this break the current functionality ? In case of current path we do still need to use "/etc/acf/service.acf" and for targetted ACF path, we need to use the path extracted from the dump entry.

oem/ibm/requester/dbus_to_file_handler.hpp
```diff
@@ -48,9 +48,11 @@ class DbusToFileHandler
     /** @brief Process the new resource dump request
      *  @param[in] vspString - vsp string
      *  @param[in] resDumpReqPass - resource dump password
+     *  @param[in] acfFilePath  ACF file path
```
jaypadath: Please change to "acfFilePath - ACF file path" in accordance with other parameter formats

jaypadath (changes requested):
Please reduce the length of the title as something like "oem-ibm: Add support for targeted ACF file"

oem/ibm/requester/dbus_to_file_handler.cpp
```diff
@@ -142,6 +143,9 @@ void DbusToFileHandler::processNewResourceDump(
             resDumpProgressIntf);
     }
 
+    info("Received acf file path {ACF_FILE_PATH} and vsp string {VSP_STRING}",
+         "ACF_FILE_PATH", acfFilePath, "VSP_STRING", vspString);
```
jaypadath: Most of the cases, this ACF file path will be empty unless its a targetted acf flow. Do we really need to trace this  or remove it ?

oem/ibm/requester/dbus_to_file_handler.cpp
Jino26: Yeah added the change to pick service.acf if we don't get path of targeted acf.

oem/ibm/requester/dbus_to_file_handler.hpp
Jino26: Done.

oem/ibm/requester/dbus_to_file_handler.cpp
Jino26: I moved it to `getAcfFileContent` function where we check for ACF file.

Pavithrab7 (approved):

manojkiraneda (approved):

PR #708, opened 2025-07-23
bios: Add a debug log trace in setDateTime command

Pavithrab7:
When a Time Reference Partition (TRP) is configured on the LPAR, PHYP sends down the SetDateTime command to BMC inorder to set the same time on the BMC. There was a debug trace in our code that told BMC that the command is sent down from PHYP.

Adding the debug trace back in the code because it would be helpful for BMC to understand if we ever got the setDateTime command from PHYP when the TRP was set.

Fixes: 724586

### DIFF
--- a/libpldmresponder/bios.cpp
+++ b/libpldmresponder/bios.cpp
@@ -167,6 +167,11 @@ Response Handler::setDateTime(const pldm_msg* request, size_t payloadLength)
         "xyz.openbmc_project.Time.Synchronization";
     constexpr auto timeSyncProperty = "TimeSyncMethod";
 
+    info("Got a setDateTime command from host to configure the BMC time");
+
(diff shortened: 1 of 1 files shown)


libpldmresponder/bios.cpp
```diff
@@ -167,6 +167,8 @@ Response Handler::setDateTime(const pldm_msg* request, size_t payloadLength)
         "xyz.openbmc_project.Time.Synchronization";
     constexpr auto timeSyncProperty = "TimeSyncMethod";
 
+    info("Got a setDateTime command from host to configure the BMC time");
```
manojkiraneda, line 170: please also add the provided phyp time  with TRP time and the [current bmc time](https://github.com/openbmc/pldm/blob/master/common/utils.cpp#L670)

manojkiraneda (approved):

PR #722, opened 2025-11-04
Build file table lazily

rahulds98:
Support for HB lid files patching
Solves an issue where when lid files are patched on host poweroff, filetable is not updated unless pldmd is restarted resulting in file transfer of wrong size Here filetable is built before hostboot

### DIFF
--- a/oem/ibm/libpldmresponder/oem_ibm_handler.hpp
+++ b/oem/ibm/libpldmresponder/oem_ibm_handler.hpp
@@ -3,6 +3,7 @@
 #include "collect_slot_vpd.hpp"
 #include "common/types.hpp"
 #include "common/utils.hpp"
+#include "file_table.hpp"
 #include "inband_code_update.hpp"
 #include "libpldmresponder/oem_handler.hpp"
 #include "libpldmresponder/pdr_utils.hpp"
@@ -169,6 +170,7 @@ class Handler : public oem_platform::Handler
                         disableWatchDogTimer();
                         startStopTimer(false);
                         pldm::responder::utils::clearLicenseStatus();
+                        pldm::filetable::clearFileTable();
                     }
                     else if (propVal ==
                              "xyz.openbmc_project.State.Host.HostState.Running")
--- a/oem/ibm/libpldmresponder/file_table.cpp
+++ b/oem/ibm/libpldmresponder/file_table.cpp
@@ -129,14 +129,21 @@ Table FileTable::operator()() const
     return table;
 }
 
+/** Single instance of filetable*/
+static FileTable fileTable;
+
 FileTable& buildFileTable(const std::string& fileTablePath)
 {
-    static FileTable table;
-    if (table.isEmpty())
+    if (fileTable.isEmpty())
     {
-        table = FileTable(fileTablePath);
+        fileTable = FileTable(fileTablePath);
     }
-    return table;
+    return fileTable;
+}
+
+void clearFileTable()
+{
+    fileTable.clear();
 }
 
 } // namespace filetable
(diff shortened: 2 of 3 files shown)


rahulds98:
Tested with:
Change has been made after checking how many times filetable is requested(3 times)
Filetable gets rebuilt on poweron which is when the filetable is requested and lid files transferred.

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
```diff
@@ -176,12 +177,19 @@ class Handler : public oem_platform::Handler
...
                     {
                         hostTransitioningToOff = true;
                     }
+                    else if (propVal ==
+                             "xyz.openbmc_project.State.Host.HostState.TransitioningToRunning")
+                    {
+                        using namespace pldm::filetable;
+                        auto& fileTable = buildFileTable(FILE_TABLE_JSON);
+                        fileTable.clear();
+                        fileTable = buildFileTable(FILE_TABLE_JSON);
```
ArchanaKakani: We can just clear the file table when host is in TransitioningToOff state.
Building the file table is already taken care as we do build the file table if it is empty (Inside buildFileTable) and for all the requests(getFileTable/readFile/writeFile/writeFileFromMemory/readFileIntoMemory) we do get the filetable so it will build the file table whenever needed. 
Do you see any issue with this?

Also can you please add comments mentioning that the change is added to support runtime HB lid patching

ArchanaKakani (changes requested):

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
manojkiraneda: +1 to what Archana mentioned, with just one suggestion - put clear logic under `xyz.openbmc_project.State.Host.HostState.Off`  rather than `TransitioningToOff state`

manojkiraneda (changes requested):

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
rahulds98: Done

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
```diff
@@ -169,19 +170,26 @@ class Handler : public oem_platform::Handler
                         disableWatchDogTimer();
                         startStopTimer(false);
                         pldm::responder::utils::clearLicenseStatus();
+                        using namespace pldm::filetable;
```
ArchanaKakani: Creating a function to clear the file table under filetable namespace will make the code look clean.

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
```diff
@@ -169,19 +170,26 @@ class Handler : public oem_platform::Handler
...
                     }
-                    else if (
-                        propVal ==
-                        "xyz.openbmc_project.State.Host.HostState.TransitioningToOff")
+                    else if (propVal ==
+                             "xyz.openbmc_project.State.Host.HostState.TransitioningToOff")
                     {
                         hostTransitioningToOff = true;
                     }
+                    else if (propVal ==
```
ArchanaKakani: Do we still need this ?

ArchanaKakani (changes requested):

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
rahulds98: No, removed

oem/ibm/libpldmresponder/oem_ibm_handler.hpp
rahulds98: Done

geissonator:
add to approvelist

ArchanaKakani (approved):

manojkiraneda (approved):

PR #724, opened 2025-11-12
oem-ibm: Clear PCIeSlot DBus Object LinkId/BusId property on topology refresh

rahulds98:
This commit clears BusIDs on all PCIE DBus Objects ensuring stale links are cleared

Testsed by enabling and disabling hb_host_usb_enablement to verify if link gets removed from bmcweb

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_pcie.cpp
+++ b/oem/ibm/libpldmresponder/file_io_type_pcie.cpp
@@ -160,6 +160,10 @@ int PCIeInfoHandler::fileAck(uint8_t /*fileStatus*/)
             }
             cables.clear();
 
+            // clears busids on all objects because only active links are
+            // updated by setTopologyAttrsOnDbus()
+            refreshAllPcieSlotBusIds();
+
             // set topology properties & host cable dbus objects
             setTopologyAttrsOnDbus();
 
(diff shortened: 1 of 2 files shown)


oem/ibm/libpldmresponder/file_io_type_pcie.cpp
```diff
@@ -1158,5 +1161,27 @@ void PCIeInfoHandler::deleteTopologyFiles()
     }
 }
 
+void PCIeInfoHandler::clearBusIDs()
+{
+    constexpr auto basePath = "/xyz/openbmc_project/inventory/system";
+    auto subTree =
+        pldm::utils::DBusHandler().getSubTreePaths(basePath, 0, {itemPCIeSlot});
```
ArchanaKakani: Changes looks good to be. 
For optimization can we use getSubTree which will also give us service hosting those object paths. And we can avoid using `setDbusProperty` directly set the dbus property here itself.
This way we can avoid multiple getservice calls made at:
https://github.com/ibm-openbmc/pldm/blob/1060/common/utils.cpp#L325

@manojkiraneda any thoughts on this?

manojkiraneda: yes , i agree. We have seen in some cases that mapper is slow changing this to getSubTree along with bus.call to set the property would be better.

rahulds98: Used getSubtree instead of getSubTreePaths and bus.call

manojkiraneda (approved):

Pavithrab7 (approved):

PR #741, opened 2026-01-11
pldm: PDR and FRU VPD additions in Huygens system

jaypadath:
This commit contains some of the below changes related to Huygens system
1. Addition of PDR folder and added the available frus for switch board and two SLEDs
2. Added state sensor effecters for memory module
3. Added new FRU VPD data for System, Chassis and Board
4. Updated FRU master json with core entity id

Tested By:
Verified most of the changes in the available Huygens simics setup

### DIFF
--- a/configurations/fru_master.json
+++ b/configurations/fru_master.json
@@ -13,6 +13,7 @@
         "xyz.openbmc_project.Inventory.Item.Vrm": 123,
         "xyz.openbmc_project.Inventory.Item.Cpu": 135,
         "xyz.openbmc_project.Inventory.Item.Bmc": 137,
+        "xyz.openbmc_project.Inventory.Item.CpuCore": 151,
         "xyz.openbmc_project.Inventory.Item.Connector": 185,
         "xyz.openbmc_project.Inventory.Item.PCIeSlot": 186,
         "xyz.openbmc_project.Inventory.Item.System": 32813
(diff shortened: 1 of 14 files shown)


configurations/events/stateSensorPdrs.json
```diff
@@ -78,7 +78,7 @@
             "stateSetId": 1,
             "event_states": [1, 3],
             "dbus": {
-                "object_path": "/xyz/openbmc_project/inventory/system/chassis/motherboard/dimm0",
+                "object_path": "/xyz/openbmc_project/inventory/system/chassis1/motherboard/dimm0",
                 "interface": "xyz.openbmc_project.State.Decorator.OperationalStatus",
```
manojkiraneda, line 82: Instead we can add support to make object_path an array.

So when we get a sensor event from remote endpoint - we can do mapper queries to both the paths listed in the array and if present we set the dbus property on both of them.

That way the code is scalable for both old and new systems. Can you remove this file from this PR - we can have a seperate PR for making this an array (since this does not break the power on)

manojkiraneda (changes requested):
@jaypadath - just remove the configurations/events/stateSensorPdrs.json file from this PR & upstream commit -  we can merge the rest of it immediately. 

The array change can be made upstream & then we can pull it downstream later since that dont block the power on.

configurations/events/stateSensorPdrs.json
jaypadath: Okay got your point. Let me remove this file as of now and re-push the change and will create sepate PR for making object_path as an array.

jaypadath:
> @jaypadath - just remove the configurations/events/stateSensorPdrs.json file from this PR & upstream commit - we can merge the rest of it immediately.
> 
> The array change can be made upstream & then we can pull it downstream later since that dont block the power on.

Done and re-pushed

manojkiraneda (approved):

PR #745, opened 2026-01-23
Fixed The Issue of Not Listing MEX CHASSIS after Restart PLDM

RameshaR45:
Mex chassis were not listing after the power on after we make power off.
and also after the restart pldm

### DIFF
--- a/host-bmc/utils.cpp
+++ b/host-bmc/utils.cpp
@@ -135,14 +135,15 @@ void addObjectPathEntityAssociations(
             }
             try
             {
-                pldm::utils::DBusHandler().getService(entity_path.c_str(),
-                                                      nullptr);
+                auto serviceName = pldm::utils::DBusHandler().getService(
+                    entity_path.c_str(), nullptr);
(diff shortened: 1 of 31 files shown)


host-bmc/utils.cpp
```diff
@@ -181,7 +185,10 @@ void addObjectPathEntityAssociations(
         try
         {
             pldm::utils::DBusHandler().getService(dbusPath.c_str(), nullptr);
```
manojkiraneda: merge both of them together

host-bmc/utils.cpp
```diff
@@ -137,12 +137,16 @@ void addObjectPathEntityAssociations(
             {
                 pldm::utils::DBusHandler().getService(entity_path.c_str(),
```
manojkiraneda: getService is duplicated here, remove one instance of it.

manojkiraneda (changes requested):

manojkiraneda (approved):

PR #753, opened 2026-03-25
host-bmc: Fix incorrect non-functional(Non Guarded) core status after concurrent code update or reboot

RameshaR45:
Fix incorrect non-functional(Non Guarded) core status after concurrent code update or reboot


RameshaR45:
due to some commit isssues

PR #757, opened 2026-04-06
oem-ibm: Redesign of systemdump to write to the BMC flash

rahulds98:
This commit implements a redesign of the system dump
handling mechanism for system dumps from HB to store
dumps directly to the BMC file system instead
of offloading to Unix Sockets without using DMA

Key changes:

1. File System Storage:
   - System dumps are now written to /tmp/dump_<fileHandle>

2. Timeout Protection:
   - Implemented timeout mechanism using sdeventplus timer
   - Timer starts on newfileavailable
   - Automatically closes file descriptor on timeout to prevent resource leaks
   - Timer stops on successful FileAck completion

3. DumpHandler Changes:
   - Modified newfileavailable() to start the timer instead of notifying dumphandler
   - Modified write() method to handle file system operations
   - Updated fileAck() to properly close file descriptor
   - Maintained original behavior for other dump types

4. Handler Timer Management:
   - Added startDumpTransferTimer() to initiate countdown
   - Added stopDumpTransferTimer() to cancel active timer
   - Added onDumpTransferTimeout() callback for timeout handling
   
   Tested By:
1. pldmtool commands to simulate writefilebytype for file type
   systemdump(PLDM_FILE_TYPE_DUMP)
2. Verified that on timeout, fd is closed and no new writes
   are accepted
3. Enforces 1 systemdump is transferred at a time

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_dump.cpp
+++ b/oem/ibm/libpldmresponder/file_io_type_dump.cpp
@@ -5,6 +5,7 @@
 #include "utils.hpp"
 #include "xyz/openbmc_project/Common/error.hpp"
 
+#include <fcntl.h>
 #include <libpldm/base.h>
 #include <libpldm/oem/ibm/file_io.h>
 #include <systemd/sd-bus.h>
@@ -17,6 +18,7 @@
 #include <cstdint>
 #include <exception>
 #include <filesystem>
+#include <format>
 #include <type_traits>
 
 PHOSPHOR_LOG2_USING;
@@ -41,6 +43,13 @@ static constexpr auto bmcDumpObjPath = "/xyz/openbmc_project/dump/bmc/entry";
 // resource dumps.
 
 int DumpHandler::fd = -1;
+std::unordered_map<FileHandle, SysDumpTransferData> DumpHandler::sysDumpMap;
+sdeventplus::Event* DumpHandler::eventLoop = nullptr;
+
+// System dump file configuration
+constexpr auto sysDumpPath = "/tmp";
+constexpr auto sysDumpFilePrefix = "sysdump_";
+
 namespace fs = std::filesystem;
 
 uint32_t DumpHandler::getDumpIdPrefix(uint16_t dumpType)
@@ -203,6 +212,11 @@ int DumpHandler::newFileAvailable(uint64_t length)
         notifyDumpType =
             sdbusplus::common::com::ibm::dump::Notify::DumpType::Resource;
     }
+    else if (dumpType == PLDM_FILE_TYPE_DUMP) // for system dump start timer and
+                                              // open the file
+    {
+        return DumpHandler::initializeSystemDumpTransfer(fileHandle);
+    }
 
     try
     {
@@ -337,10 +351,33 @@ void DumpHandler::writeFromMemory(uint32_t, uint32_t length, uint64_t address,
                              sharedAIORespDataobj, event);
 }
 
-int DumpHandler::write(const char* buffer, uint32_t, uint32_t& length,
+int DumpHandler::write(const char* buffer, uint32_t offset, uint32_t& length,
                        oem_platform::Handler* /*oemPlatformHandler*/,
                        struct fileack_status_metadata& /*metaDataObj*/)
 {
+    // Write to file for PLDM_FILE_TYPE_DUMP
+    if (dumpType == PLDM_FILE_TYPE_DUMP)
+    {
+        // Check if transfer exists for this fileHandle
+        if (!sysDumpMap.contains(fileHandle))
+        {
+            error(
+                "Write rejected for system dump: no active transfer for fileHandle {HANDLE}",
+                "HANDLE", fileHandle);
+            return PLDM_ERROR;
+        }
+        int dumpFd = sysDumpMap.at(fileHandle).fd;
+        int rc = ::pwrite(dumpFd, buffer, length, offset);
+        if (rc < 0)
+        {
+            error("Failed to write to dump file, with error:{ERRNO}", "ERRNO",
+                  errno);
+            return PLDM_ERROR;
+        }
+        length = rc; // Update length with actual bytes written
+        return PLDM_SUCCESS;
+    }
+
     int rc = writeToUnixSocket(DumpHandler::fd, buffer, length);
     if (rc < 0)
     {
@@ -359,6 +396,20 @@ int DumpHandler::write(const char* buffer, uint32_t, uint32_t& length,
 
 int DumpHandler::fileAck(uint8_t fileStatus)
 {
+    if (dumpType == PLDM_FILE_TYPE_DUMP) // for system dump
+    {
+        // Check if the fileHandle exists in the map
+        if (!sysDumpMap.contains(fileHandle))
+        {
+            error(
+                "System dump transfer session does not exist for fileHandle {FILE_HANDLE}",
+                "FILE_HANDLE", fileHandle);
+            return PLDM_INVALID_FILE_HANDLE;
+        }
+        sysDumpMap.erase(fileHandle);
+        return PLDM_SUCCESS;
+    }
+
     auto path = findDumpObjPath(fileHandle);
     if (dumpType == PLDM_FILE_TYPE_RESOURCE_DUMP_PARMS)
     {
@@ -444,8 +495,7 @@ int DumpHandler::fileAck(uint8_t fileStatus)
             return PLDM_SUCCESS;
         }
 
-        if (dumpType == PLDM_FILE_TYPE_DUMP ||
-            dumpType == PLDM_FILE_TYPE_RESOURCE_DUMP)
+        if (dumpType == PLDM_FILE_TYPE_RESOURCE_DUMP)
         {
             PropertyValue value{true};
             DBusMapping dbusMapping{path, dumpEntry, "Offloaded", "bool"};
@@ -784,5 +834,58 @@ int DumpHandler::fileAckWithMetaData(
 
     return PLDM_ERROR;
 }
+
+int DumpHandler::initializeSystemDumpTransfer(FileHandle fileHandle)
+{
+    // Check if transfer already exists for this fileHandle
+    if (sysDumpMap.contains(fileHandle))
+    {
+        warning(
+            "Dump transfer already active for fileHandle {FILE_HANDLE}, cleaning up old transfer",
+            "FILE_HANDLE", fileHandle);
+        sysDumpMap.erase(fileHandle);
+    }
+
+    // Open file for writing system dump
+    std::string dumpFilePath =
+        std::format("{}/{}{}", sysDumpPath, sysDumpFilePrefix, fileHandle);
+    int dumpFd = open(dumpFilePath.c_str(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
+    if (dumpFd < 0)
+    {
+        error("Failed to open dump file {PATH}, with error: {ERRNO}", "PATH",
+              dumpFilePath, "ERRNO", errno);
+        return PLDM_ERROR;
+    }
+    // Create timer, start it, and add to map with fd
+    try
+    {
+        auto timer = std::make_unique<SysDumpTimer>(
+            *DumpHandler::eventLoop, [fileHandle](SysDumpTimer&) {
+                DumpHandler::onDumpTransferTimeout(fileHandle);
+            });
+        timer->restart(
+            std::chrono::minutes(DumpHandler::sysDumpTimeoutMinutes));
+        sysDumpMap.emplace(fileHandle,
+                           SysDumpTransferData(std::move(timer), dumpFd));
+    }
+    catch (const std::exception& e)
+    {
+        error(
+            "Failed to create dump transfer for fileHandle {FILE_HANDLE}: {ERROR}",
+            "FILE_HANDLE", fileHandle, "ERROR", e.what());
+        close(dumpFd);
+        return PLDM_ERROR;
+    }
+    return PLDM_SUCCESS;
+}
+
+void DumpHandler::onDumpTransferTimeout(FileHandle fileHandle)
+{
+    error(
+        "System dump transfer aborted due to timeout for fileHandle {FILEHANDLE}. File may be incomplete.",
+        "FILEHANDLE", fileHandle);
+    sysDumpMap.erase(fileHandle);
+}
+
 } // namespace responder
 } // namespace pldm
--- a/oem/ibm/libpldmresponder/file_io_type_dump.hpp
+++ b/oem/ibm/libpldmresponder/file_io_type_dump.hpp
@@ -2,11 +2,66 @@
 
 #include "file_io_by_type.hpp"
 
+#include <sdeventplus/event.hpp>
+#include <sdeventplus/utility/timer.hpp>
+
+#include <memory>
+#include <unordered_map>
+#include <utility>
+
 namespace pldm
 {
 namespace responder
 {
 using DumpEntryInterface = std::string;
+using FileHandle = uint32_t;
+using SysDumpTimer =
+    sdeventplus::utility::Timer<sdeventplus::ClockId::Monotonic>;
+
+/** @struct SysDumpTransferData
+ *
+ *  @brief Wrapper for systemdump that owns timer and fd
+ */
+struct SysDumpTransferData
+{
+    std::unique_ptr<SysDumpTimer> timer;
+    int fd;
+
+    SysDumpTransferData(std::unique_ptr<SysDumpTimer>&& t, int f) :
+        timer(std::move(t)), fd(f)
+    {}
+
+    ~SysDumpTransferData()
+    {
+        if (fd >= 0)
+        {
+            close(fd);
+        }
+    }
+
+    // Prevent copying
+    SysDumpTransferData(const SysDumpTransferData&) = delete;
+    SysDumpTransferData& operator=(const SysDumpTransferData&) = delete;
+    SysDumpTransferData(SysDumpTransferData&& other) noexcept :
+        timer(std::move(other.timer)), fd(other.fd)
+    {
+        other.fd = -1;
+    }
+    SysDumpTransferData& operator=(SysDumpTransferData&& other) noexcept
+    {
+        if (this != &other)
+        {
+            if (fd >= 0)
+            {
+                close(fd);
+            }
+            timer = std::move(other.timer);
+            fd = other.fd;
+            other.fd = -1;
+        }
+        return *this;
+    }
+};
 
 /** @class DumpHandler
  *
@@ -16,6 +71,9 @@ using DumpEntryInterface = std::string;
 class DumpHandler : public FileHandler
 {
   public:
+    // System dump transfer timeout in minutes
+    static constexpr auto sysDumpTimeoutMinutes = 15;
+
     /** @brief DumpHandler constructor
      */
     DumpHandler(uint32_t fileHandle, uint16_t fileType) :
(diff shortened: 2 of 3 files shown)


oem/ibm/libpldmresponder/oem_ibm_handler.cpp
```diff
@@ -2128,6 +2129,70 @@ void pldm::responder::oem_ibm_platform::Handler::triggerHostEffecter(
...
 }
 
+void Handler::startDumpTransferTimer()
+{
+    constexpr auto timeoutMinutes = DUMP_TIMEOUT_MINUTES;
+    uint32_t fileHandle = DumpHandler::activeSysDumpFileHandle;
+    if (dumpTransferTimer)
+    {
+        info(
+            "Restarting system dump transfer timer for fileHandle {FILE_HANDLE}",
```
jaypadath: Will the Starting/Restarting and Stop traces flood the journal ?

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -22,6 +22,9 @@ class DumpHandler : public FileHandler
         FileHandler(fileHandle), dumpType(fileType)
     {}
 
+    // Allow oem_ibm_platform::Handler to access private members for timeout
+    friend class pldm::responder::oem_ibm_platform::Handler;
```
jaypadath: Currently we are only using friend classes for Test purpose in pldm. Is there any way we could avoid this ? We do use "oem_ibm_platform::Handler" in some of scenarios like collect vpd, lid handling and in band code update. Yea if there is no one other option we can go with friend class.

jaypadath (changes requested):

oem/ibm/libpldmresponder/oem_ibm_handler.cpp
rahulds98: This trace is once per system dump, the timer is only started on newfileavailable, so this should be fine, agree?

jaypadath:
Also please reduce the title length "oem-ibm: Redesign the system dump offload" or something similar.

"This commit implements a redesign of the system dump handling mechanism to for non DMA system dumps from HB to store dumps directly to the BMC file system instead of offloading directly to Unix Sockets" --> This is little confusing and please re-phrase something similar to
"This commit implements a redesign of the system dump handling mechanism.  System dumps from HB would be offloaded directly to the BMC file system instead of offloading directly to Unix Sockets without using DMA." 

Also add a "Tested By" field in the commit message and mentioned some of the use cases tested.

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: I have removed the friend class, replaced it with a few static getters.

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -337,10 +360,63 @@ void DumpHandler::writeFromMemory(uint32_t, uint32_t length, uint64_t address,
...
+            error(
+                "Rejecting write for fileHandle {FILEHANDLE} - no active transfer session (active={ACTIVE})",
+                "FILEHANDLE", fileHandle, "ACTIVE", activeSysDumpFileHandle);
+            return PLDM_ERROR;
+        }
+
+        // Open file only on first write (offset==0) or if fd is invalid
+        if (offset == 0 && DumpHandler::fd < 0)
+        {
+            std::string dumpFilePath =
```
manojkiraneda: use `std::format` to format arbitrary strings like these.

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -41,6 +42,13 @@ static constexpr auto bmcDumpObjPath = "/xyz/openbmc_project/dump/bmc/entry";
 // resource dumps.
 
 int DumpHandler::fd = -1;
+oem_platform::Handler* DumpHandler::oemHandler = nullptr;
+uint32_t DumpHandler::activeSysDumpFileHandle = 0;
```
manojkiraneda: why are we tracking the activedump ? we create an object for every dump, if we start the timer within the object itself then we can have multiple dumps with multiple timers running - and all of them can happen parallely.

meson.build
```diff
@@ -112,6 +112,10 @@ if get_option('oem-ibm').allowed()
         '/usr/local/share/hostfw/alternate',
     )
     conf_data.set('DMA_MAXSIZE', get_option('oem-ibm-dma-maxsize'))
+    conf_data.set(
+        'DUMP_TIMEOUT_MINUTES',
+        get_option('oem-ibm-dump-timeout-minutes'),
+    )
```
manojkiraneda: I think a meson option for this is too much to cary for something which is only used by IBM. I agree that we did similar things before, but now it left us with ton of meson options which does not really change.

So for this let's get it hardcoded into the code itself & drop the meson config all together.

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -203,6 +211,21 @@ int DumpHandler::newFileAvailable(uint64_t length)
...
+    {
+        // Start system dump transfer timer and mark transfer as active
+        activeSysDumpFileHandle = fileHandle;
+        if (oemHandler != nullptr)
+        {
+            auto oemIbmHandler =
+                dynamic_cast<oem_ibm_platform::Handler*>(oemHandler);
+            if (oemIbmHandler != nullptr)
+            {
+                oemIbmHandler->startDumpTransferTimer();
```
manojkiraneda: we need the timer to be part of the dump object itself, rather than it being an object under ibm handler which has controls from the dump object. that way you can eliminate the need for oemIBMHandler and also needing to track an active system dump.

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -359,6 +435,50 @@ int DumpHandler::write(const char* buffer, uint32_t, uint32_t& length,
...
+            {
+                info(
+                    "System dump transfer complete: {BYTES} bytes written for fileHandle={FILEHANDLE}",
+                    "BYTES", static_cast<uint64_t>(fileSize), "FILEHANDLE",
+                    fileHandle);
+            }
+            close(DumpHandler::fd);
+            DumpHandler::fd = -1;
+
+            std::string dumpFilePath =
```
manojkiraneda: std::format here as well.

manojkiraneda (changes requested):

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: done

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: Modified to add timer logic within dump handler itself, please note that it is sttic since dump handler gets recreated for each dump fileio command.

meson.build
rahulds98: removed meson and added timeout in code

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: Modified to add a static map of dump handles and their timers

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: done

geissonator:
Clang format issues in CI - https://sys-openbmc-dev-jenkins.swg-devops.com/job/OpenBMC-Dev/job/build/job/github-public/job/repo-ci/job/public-ibm-openbmc-downstream-repo-ci-pr-pldm/1930/console

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +73,57 @@ class DumpHandler : public FileHandler
...
+    /** @brief Set event loop for timer management
+     *  @param[in] event - Event loop pointer
+     */
+    static void setEventLoop(sdeventplus::Event* event)
+    {
+        if (eventLoop == nullptr)
+        {
+            eventLoop = event;
+        }
+    }
```
manojkiraneda: just initialize the event object in the constructor itself ? by passing it from main() ?

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +73,57 @@ class DumpHandler : public FileHandler
...
+
+    // Pair: <timer, fd> for each active system dump transfer
+    static std::unordered_map<
+        uint32_t, std::pair<std::unique_ptr<sdeventplus::utility::Timer<
+                                sdeventplus::ClockId::Monotonic>>,
+                            int>>
+        sysDumpMap; //!< Map of fileHandle to (timer, fd) for system dumps
+
+    static sdeventplus::Event* eventLoop; //!< Event loop for timer management
+    uint16_t dumpType;                    //!< type of the dump
```
manojkiraneda: this is just removed and being added ..move it back to its place.

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +73,57 @@ class DumpHandler : public FileHandler
...
     ~DumpHandler() {}
 
   private:
-    static int fd;             //!< fd to manage the dump offload to bmc
-    uint16_t dumpType;         //!< type of the dump
+    static int fd; //!< fd to manage the dump offload to bmc
+
+    // Pair: <timer, fd> for each active system dump transfer
+    static std::unordered_map<
+        uint32_t, std::pair<std::unique_ptr<sdeventplus::utility::Timer<
```
manojkiraneda: use type definitions rather than using plain uint32_t e.t.c

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -203,6 +216,25 @@ int DumpHandler::newFileAvailable(uint64_t length)
...
+        int dumpFd =
+            open(dumpFilePath.c_str(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
+        if (dumpFd < 0)
+        {
+            error("Failed to open dump file {PATH}, with error: {ERRNO}",
+                  "PATH", dumpFilePath, "ERRNO", errno);
+            return PLDM_ERROR;
+        }
+        // Create timer and store both timer and fd in map
+        startDumpTransferTimer(fileHandle);
```
manojkiraneda: error handling ? if this fails we will have dumpFd leaked ?

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -203,6 +216,25 @@ int DumpHandler::newFileAvailable(uint64_t length)
...
+        sysDumpMap[fileHandle].second = dumpFd;
```
manojkiraneda: if fileHandle already exists, then we override it with a new fd without closing the old one ? this might leak the fd as well right ?

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -784,5 +860,68 @@ int DumpHandler::fileAckWithMetaData(
...
+}
+
+void DumpHandler::onDumpTransferTimeout(uint32_t fileHandle)
+{
+    // Close the fd and remove from map
+    closeDumpFile(fileHandle);
+    error(
+        "System dump transfer aborted due to timeout for fileHandle {FILEHANDLE}. File may be incomplete.",
+        "FILEHANDLE", fileHandle);
+}
```
manojkiraneda: this all looks a little dangerous rahul , since we have multiple paths to leak fd's and also multiple erase paths. I think we should not look at fd & timer as two different entities. I don't think we want one of it to exist without another one right ? 

So I think the cleanest way to handle this is by using a RAII helper struct/class that holds a unique pointer to the timer instance & the destructor of that object should automatically close the fd. That way every time we remove an entry from the map we close the fd.

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: I think this would require modifying the constructortor and modifying the gethandlerbytype() call for all dump types which I think is not ideal

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: Done

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: Added try catch here to only open fd if timer starts

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: Added a RAII struct that couples fd and timer and fized all leak paths

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -359,6 +397,21 @@ int DumpHandler::write(const char* buffer, uint32_t, uint32_t& length,
 
 int DumpHandler::fileAck(uint8_t fileStatus)
 {
+    if (dumpType == PLDM_FILE_TYPE_DUMP) // for system dump
+    {
+        // Check if the fileHandle exists in the map
+        auto entry = sysDumpMap.find(fileHandle);
+        if (entry == sysDumpMap.end())
```
manojkiraneda: you could use `map.contains()` instead - thats much more readable.

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -784,5 +836,59 @@ int DumpHandler::fileAckWithMetaData(
 
     return PLDM_ERROR;
 }
+
+int DumpHandler::initializeSystemDumpTransfer(FileHandle fileHandle)
+{
+    // Check if transfer already exists for this fileHandle
+    auto entry = sysDumpMap.find(fileHandle);
+    if (entry != sysDumpMap.end())
```
manojkiraneda: map.contains() here as well

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +121,48 @@ class DumpHandler : public FileHandler
...
+    static std::unordered_map<FileHandle, SysDumpTransfer>
+        sysDumpMap; //!< Map of fileHandle to SysDumpTransfer (timer + fd) for
+                    //!< system dumps
+    static sdeventplus::Event* eventLoop; //!< Event loop for timer management
     std::string
-        resDumpRequestDirPath; //!< directory where the resource
+        resDumpRequestDirPath;            //!< directory where the resource
                                //!< dump request parameter file is stored
-    int unixFd;                //!< fd to temporarily hold the fd created.
+    int unixFd; //!< fd to temporarily hold the fd created.
```
manojkiraneda: I dont see this being used anywhere ? is it required ?

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +121,48 @@ class DumpHandler : public FileHandler
...
+
     /** @brief DumpHandler destructor
      */
     ~DumpHandler() {}
 
   private:
-    static int fd;             //!< fd to manage the dump offload to bmc
-    uint16_t dumpType;         //!< type of the dump
+    static int fd;     //!< fd to manage the dump offload to bmc
+    uint16_t dumpType; //!< type of the dump
```
manojkiraneda: this is till showing as diff, you might wanna revert this changes back to original since that is not being added as art of this commit.

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +121,48 @@ class DumpHandler : public FileHandler
...
+        eventLoop = &event;
+    }
+
+    /** @brief Get event loop reference
+     *  @return Reference to the event loop
+     */
+    static sdeventplus::Event& getEventLoop()
+    {
+        return *eventLoop;
+    }
```
manojkiraneda: do we really need this getter ? this function is called in `initializeSystemDumpTransfer` which is a public function. `initializeSystemDumpTransfer` should be able to directly access its private varible eventloop without needing a getter right ?

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -63,17 +121,48 @@ class DumpHandler : public FileHandler
...
-    uint16_t dumpType;         //!< type of the dump
+    static int fd;     //!< fd to manage the dump offload to bmc
+    uint16_t dumpType; //!< type of the dump
+    static std::unordered_map<FileHandle, SysDumpTransfer>
+        sysDumpMap; //!< Map of fileHandle to SysDumpTransfer (timer + fd) for
+                    //!< system dumps
+    static sdeventplus::Event* eventLoop; //!< Event loop for timer management
     std::string
-        resDumpRequestDirPath; //!< directory where the resource
+        resDumpRequestDirPath;            //!< directory where the resource
```
manojkiraneda: this change is also not related - you seem to add some spaces here. revert it back to original.

oem/ibm/libpldmresponder/file_io_type_dump.hpp
```diff
@@ -36,7 +94,7 @@ class DumpHandler : public FileHandler
                      oem_platform::Handler* /*oemPlatformHandler*/);
 
     virtual int write(const char* buffer, uint32_t offset, uint32_t& length,
-                      oem_platform::Handler* /*oemPlatformHandler*/,
+                      oem_platform::Handler* oemPlatformHandler,
```
manojkiraneda: i think we can move it back to original since we are not using it anymore  ?

manojkiraneda (changes requested):

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: I think we can clean this up after we modify writefrommemory() for DMA, since its used in that function for now.

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: The clang formatting seems to have added this here, will revert

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: Done

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: Removed this function

oem/ibm/libpldmresponder/file_io_type_dump.hpp
rahulds98: Done

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: done

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: done

manojkiraneda (approved):

PR #761, opened 2026-04-28
pldm: Ignore timeout errors during activation

ArchanaKakani:


oem/ibm/libpldmresponder/inband_code_update.cpp
```diff
@@ -165,10 +165,32 @@ int CodeUpdate::setRequestedActivation()
...
+            error(
+                "Failed to set property {PROPERTY} at path '{PATH}' and interface "
+                "'{INTERFACE}', error - {ERROR}",
+                "PATH", dbusMapping.objectPath, "INTERFACE", dbusMapping.interface,
+                "PROPERTY", dbusMapping.propertyName, "ERROR", e);
+            rc = PLDM_ERROR;
+        }
+        else
+        {
+            info("RequestedActivation at path '{PATH}' is ignored", "PATH", dbusMapping.objectPath);
```
anoo1: Could probably say that the Timeout error is ignored, since the msg may be misinterpreted to be that pldm is not going to start the update and by ignoring to set the RequestedActivation property.

anoo1 (approved):

oem/ibm/libpldmresponder/inband_code_update.cpp
ArchanaKakani: Updated the message

anoo1 (approved):

PR #770, opened 2026-06-23
oem-ibm: Fix out-of-bounds vector access in SMS menu

jaypadath:
This commit fixes the out-of-bounds vector access in SMS menu. Below changes are made:
1. Added bounds checking
2. Fixed the indexing
3. Returned error on empty lengths

Tested By:
Tested the changes in one of balcones systems and operation was successful without pldm crash

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_smsmenu.cpp
+++ b/oem/ibm/libpldmresponder/file_io_type_smsmenu.cpp
@@ -21,12 +21,44 @@ uint32_t SmsMenuHandler::readVecContent(
 {
     uint32_t size = 0;
     constexpr auto bitPos = 8;
-    std::vector<char> userPassLenArr(inputVec.begin() + startIdx,
-                                     inputVec.begin() + endIdx);
-    for (uint32_t idx = 0; idx < bitPos; idx++)
+    constexpr auto bytesPerUint32 = sizeof(uint32_t);
+
+    // Validate indices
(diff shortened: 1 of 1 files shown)


oem/ibm/libpldmresponder/file_io_type_smsmenu.cpp
```diff
@@ -21,12 +21,32 @@ uint32_t SmsMenuHandler::readVecContent(
...
+    if (endIdx > inputVec.size())
     {
-        size |= (uint32_t)userPassLenArr[idx] << bitPos * idx;
+        error("endIdx {END} exceeds vector size {SIZE}", "END", endIdx, "SIZE",
+              inputVec.size());
+        return 0;
     }
+
+    // Read 4 bytes to construct uint32_t
+    for (uint32_t idx = 0; idx < bytesPerUint32; idx++)
```
manojkiraneda, line 55: i am scared to look at patterns like these jay - these sort of things are what causes crashes and very error prone.

manojkiraneda (approved):
Since this is a stop ship and this changes are tested fine, I am okay with the changes made. But i think this entire feature function would need a refactor (with much safer c++ api's).  or worst case we need to have test API's with various inputs to see if we are safe.

oem/ibm/libpldmresponder/file_io_type_smsmenu.cpp
jaypadath: Yea we need to have better way to handle this

PR #775, opened 2026-07-02
oem-ibm: Increase bmcweb connect wait for dump

ArchanaKakani:
We are facing issues with dump offload.
BMCWeb not connecting to the pldm socket in time causing PLDM to close the socket used to transfer dump data.
Increasing the wait for the connection from Bmcweb.


jaypadath:
if (retval == 0)
    {
        error(
            "Timed out while waiting for connection on unix socket from external module");
        close(sock);
}

Above change is missing compared to the original PR https://github.com/ibm-openbmc/pldm/pull/682.

Would it be good to add here ?

PR #777, opened 2026-07-06
oem-ibm: enforce only one system dump transfer

rahulds98:
Replaced the sysDumpMap (unordered_map) with a single optional slot (sysDumpHandle + sysDumpTransfer) to enforce that only one system dump transfer can be active at a time.

- initializeSystemDumpTransfer: reject any new request while a transfer is already active; distinguish duplicate same-handle (warning) vs conflicting different-handle (error)
- write/fileAck: guard against mismatched fileHandle
- onDumpTransferTimeout: reset the single slot instead of map erase
- Increase timeout for system dump transfer to 60 minutes

### DIFF
--- a/oem/ibm/libpldmresponder/file_io_type_dump.cpp
+++ b/oem/ibm/libpldmresponder/file_io_type_dump.cpp
@@ -43,7 +43,8 @@ static constexpr auto bmcDumpObjPath = "/xyz/openbmc_project/dump/bmc/entry";
 // resource dumps.
 
 int DumpHandler::fd = -1;
-std::unordered_map<FileHandle, SysDumpTransferData> DumpHandler::sysDumpMap;
+std::optional<FileHandle> DumpHandler::sysDumpHandle;
+std::unique_ptr<SysDumpTransferData> DumpHandler::sysDumpTransfer;
 sdeventplus::Event* DumpHandler::eventLoop = nullptr;
 
 // System dump file configuration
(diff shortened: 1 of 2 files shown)


Pavithrab7:
Can you also add how this change was tested in the commit section?

Pavithrab7 (changes requested):
Already left a comment above

oem/ibm/libpldmresponder/file_io_type_dump.cpp
```diff
@@ -837,13 +838,23 @@ int DumpHandler::fileAckWithMetaData(
...
-        warning(
-            "Dump transfer already active for fileHandle {FILE_HANDLE}, cleaning up old transfer",
-            "FILE_HANDLE", fileHandle);
-        sysDumpMap.erase(fileHandle);
+        if (*sysDumpHandle == fileHandle)
+        {
+            warning(
+                "Duplicate newFileAvailable for same fileHandle {FILE_HANDLE}, ignoring",
+                "FILE_HANDLE", fileHandle);
+        }
```
Pavithrab7: here in both the cases we want to return PLDM_ERROR right?

Pavithrab7 (approved):

oem/ibm/libpldmresponder/file_io_type_dump.cpp
rahulds98: correct, just logging different levels in each case

ArchanaKakani:
> Replaced the sysDumpMap (unordered_map) with a single optional slot (sysDumpHandle + sysDumpTransfer) to enforce that only one system dump transfer can be active at a time.
> 
> * initializeSystemDumpTransfer: reject any new request while a transfer is already active; distinguish duplicate same-handle (warning) vs conflicting different-handle (error)
> * write/fileAck: guard against mismatched fileHandle
> * onDumpTransferTimeout: reset the single slot instead of map erase
> * Increase timeout for system dump transfer to 60 minutes

===>write/fileAck: guard against mismatched fileHandle
As we are supporting only one system dump at a time do we really get the mismatch filehandle in write/fileAck?
Could you explain the case where this can happen?

====>Increase timeout for system dump transfer to 60 minutes
Could you provide the details behind choosing the 60 minute timeout

=====>distinguish duplicate same-handle (warning) vs conflicting different-handle (error)
Do we really need to distinguish these cases? Do we have separate handling for them?

rahulds98:
> ===>write/fileAck: guard against mismatched fileHandle
As we are supporting only one system dump at a time do we really get the mismatch filehandle in write/fileAck?
Could you explain the case where this can happen?

If a write comes in for a different filehandle than the one the newfileavailable was sent for, rejecting it makes sense, right?

> ====>Increase timeout for system dump transfer to 60 minutes
Could you provide the details behind choosing the 60 minute timeout

This was part of a discussion which concluded that user initiated dumps which could go upto 15GB could take longer because of write, we can reduce it later once HB starts using DMA

> =====>distinguish duplicate same-handle (warning) vs conflicting different-handle (error)
Do we really need to distinguish these cases? Do we have separate handling for them?

Fixed this, we reject either case, log a single trace

ArchanaKakani (approved):

manojkiraneda (approved):

PR #780, opened 2026-07-13
host-bmc: Support of Availability state set

jaypadath:
This commit adds the support for setting the Availability state set or the dbus Enable property values when requested from host.

Tested By:
Sanity verification done in Huygens simics setup.

### DIFF
--- a/host-bmc/host_pdr_handler.cpp
+++ b/host-bmc/host_pdr_handler.cpp
@@ -324,6 +324,11 @@ int HostPDRHandler::handleStateSensorEvent(
 
             break;
         }
+        else if (stateSetId[0] == PLDM_STATE_SET_AVAILABILITY)
+        {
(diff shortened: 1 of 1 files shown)


host-bmc/host_pdr_handler.cpp
```diff
@@ -1846,6 +1856,15 @@ void HostPDRHandler::setOperationStatus()
                         stateSetIds[0]);
                     return;
                 }
+                else if (stateSetId[0] == PLDM_STATE_SET_AVAILABILITY)
```
Pavithrab7: is it not stateSetIds[0] ?
because below we see stateSetIds[0] in line 1868

jaypadath: @Pavithrab7 I have updated and re-pushed

Pavithrab7 (approved):

ArchanaKakani (approved):

PR #793, opened 2026-08-26
oem-ibm: change state set for pgood sensors/effecters

rahulds98:
Update pgood state set in sensor (4.json) and effecter (11.json) PDRs for the Huygens platform:

- Change state set id from 257 to 260
- Update states from [1, 4] to [1, 10] (1=on, 10=hard off)
- Change property_type from bool to int32
- Update property_values from [true, false] to [1, 0]
- Remove duplicate pgood effecter entries for chassis1-chassis8, pgood is read only
Tested:

used pldmtool to verify sensor/effecter entries


jaypadath (approved):

Pavithrab7 (changes requested):
I do not see any changes in the 11.json, I just see a removal of the pgood effecters, do we not have to model the effecters? 
is it because its a read-only now?

rahulds98:
> I do not see any changes in the 11.json, I just see a removal of the pgood effecters, do we not have to model the effecters?
is it because its a read-only now?

Yes, setting the state on these effecters does not really change the property since it's read only and it was also agreed to remove these effecters because PHYP or any other clients will not be able to power off the sleds

Pavithrab7 (approved):

PR #800, opened 2026-09-07
oem-ibm: Add new FRU vpd keywords

jaypadath:
Below new vpd keywords are added under VCEN and VSYS vpd records.

VCEN: Z2
VSYS: AA, AB, AK, AT, BA, BN and MM

Tested By:
Verified the change in Huygens simics setup


Pavithrab7 (commented):
Jay, are these FRU field types, alphabetical in order? Does anything change if added in the middle?

Also, where can we see what these Property name corresponds to?

PR #802, opened 2026-09-09
bios: Validate persisted BIOS attributes before restore

ArchanaKakani:
Discard persisted BIOS attribute values that are invalid for the current firmware level. Previously, persisted values were restored without validation for integer and string attributes, while invalid enum values caused an error and prevented the attribute from being added to the table.

This could result in multiple unrecoverable errors being logged during IPL when BIOS attribute details changed between firmware versions.

Validate persisted values for integer, string, and enum attributes before restoring them during power-on. If validation fails, restore the attribute with its default value instead


ArchanaKakani:
These changes are taken in another commit

